PR #110, opened 2016-02-18
Mqtt version info

JuergenKosel:
Hello,

please pull in these changes, which allow that make sets the version information into a generated file,
which is not tracked by git.
Doing this way, the sources tracked by source control do not need to be reverted after running make.

Greetings
   Juergen


icraggs:
Thanks Juergen.  I will get to looking at this and your other C client changes soon.

PR #116, opened 2016-04-06
Reorganize-samples-in-sets

abhatikar:
Grouped the samples for the library into sets which have publishers and subscribers paired accordingly so that a newbie can run it without any confusion.

Signed-off-by: Abhijeet Bhatikar abhijeet4jobs@gmail.com


icraggs:
Hello.  Wouldn't it be better to have the set directory names be meaningful?  I am renaming some of them to be better named - stdinpub becomes paho_c_pub for instance.  Some of the samples come from the documentation, so their subdirectory could be called "doc" or something similar.

abhatikar:
I can do that, could you please tell me any particular names in your mind, I could make the changes.
Thanks.

icraggs:
Thanks for the suggestion and PR.  I've changed the names of the samples to be more consistent and show the relationships between them.

PR #122, opened 2016-04-20
Fix mem leak

guilhermeferreira:
This branch has one single commit that fix a memory leak on Socket_new(). The leak occurs because the result pointer, returned by getaddrinfo(3), is lost, updated by res. The result pointer should be kept unchanged, so the freeaddrinfo(3) is able to release all the items from the list.


ralight:
You appear to have brought along an extra 37 commits unrelated to your fix. Could you please try to update the pull request to only have your commit?

guilhermeferreira:
Ok, done. Thanks!

PR #125, opened 2016-05-05
Build cmake fix

guilhermeferreira:
These commits fix two small things in the CMake scripts:
1. Header files MQTTAsync.h, MQTTClient.h and MQTTClientPersistence.h should be copied to include directory on "make install" command. This behavior exists in the regular Makefile, but were missing from CMakeLists.txt;
2. The sample application, stdinpuba, was not in the compilation;


icraggs:
Thanks!

PR #126, opened 2016-05-09
remove compiler warning for unused variable

stefboerrigter:
remove compiler warning for unused but set variable


JuergenKosel:
Why do you disable the code with "#if 0" ?
Removing dead code is usually the better choice.

stefboerrigter:
i changed my commit a bit, i agree it should (could) be removed, but the
other code that was using it used also #if 0..
but we have version history :)

2016-05-09 16:37 GMT+02:00 Juergen Kosel notifications@github.com:

PR #130, opened 2016-05-17
add padding and alignment

stefboerrigter:
i found that padding might have an issue when compiling for x64 machines with x32 library or vice versa, this way it is handled

### DIFF
--- a/src/MQTTClient.h
+++ b/src/MQTTClient.h
@@ -205,7 +205,7 @@ typedef int MQTTClient_token;
  * MQTTClient_publish(), MQTTClient_receive(), MQTTClient_freeMessage()
  * and MQTTClient_messageArrived()).
  */
-typedef struct
+typedef struct __attribute__((packed, aligned(4)))
 {
 	/** The eyecatcher for this structure.  must be MQTM. */
 	char struct_id[4];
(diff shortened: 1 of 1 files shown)


ralight:
Does this maintain platform compatibility?

If this is being done for MQTTClient.h then I'm sure the same applies to the async client as well.

Finally, is there any chance that this will break ABI compatibility?

stefboerrigter:
I guess so,
Not sure, can you test?
Op 18 mei 2016 11:39 schreef "Roger Light" notifications@github.com:

icraggs:
"i found that padding might have an issue when compiling for x64 machines with x32 library or vice versa, this way it is handled"

What is the issue that you want to fix  (building a 64-bit application with a 32 bit Paho library)? This won't work on Windows as it stands.  It may not work on the Mac. (We're working on automated build tests for pull requests).

stefboerrigter:
the issue i fixed with this for my own 'build' whas building 64bit
application with 32 bit paho library (or vice versa).
when padding is different or sizes for pointers are off it can behave quite
different.
perhaps it's just input for a discussion on your side?

2016-05-18 13:12 GMT+02:00 Ian Craggs notifications@github.com:

icraggs:
Replaced by issue https://github.com/eclipse/paho.mqtt.c/issues/138

PR #131, opened 2016-05-20
Change build system from GNU Make to Autotools

guilhermeferreira:
---
##  Advantages
1. Choose whether to build static, dynamic or both kinds of libraries;
2. Choose whether to build synchronous (MQTTClient_xxx), asynchronous
   (MQTTAsync_xxx), or both types of libraries;
3. Choose whether or not to build SSL-aware libraries;
4. Choose whether or not to build documentation, test or samples;
5. Link the samples statically with the library, so the library can
   be profiled;
6. Easier cross compilation;
7. Automatically find the required libraries to link (pthread,
   libcrypto and libssl);
8. Automatic handling of portability issues. For example, library
   naming differs accross OS.

---
##  Build instructions

The first step is to run the bootstrap script:

  $ ./bootstrap

Next, in order to build only the static library:

  $ ./configure --disable-shared --enable-static

The command above builds:

  libpaho-mqtt3a.a
  libpaho-mqtt3c.a

Or, in order to build only the dynamic library:

  $ ./configure --enable-shared --disable-static

The previous command builds:

  libpaho-mqtt3a.so.1.0.3
  libpaho-mqtt3c.so.1.0.3

and their respective links and sonames.

In order to build static only and link the test samples to static,
so it can be profiled, issue the following command:

 $ ./configure --enable-samples --disable-shared \
   CFLAGS=-pg LDFLAGS=-pg

In order to build the SSL-aware libraries, use the --with-ssl option:

  $ ./configure --disable-shared --enable-static --with-ssl

which builds the synchronous and asynchronous secure versions:

  libpaho-mqtt3a.a
  libpaho-mqtt3as.a
  libpaho-mqtt3c.a
  libpaho-mqtt3cs.a

Also, there is an option to enable/disable high level of warnings
(--enable-peak-warnings).

When done with the configuration, build the project:

  $ make

---
##  Remarks

The regular Make version is around 2.5 times faster than the Autotools
version. The overall time for Autotools include the time of bootstrap,
configure and make. The regular Make time include only the make command.

Successfully built on and for:
- Ubuntu 10.04 (x86 32-bit) with gcc 4.4.3 and GNU libc 2.11.1;
- Ubuntu 10.04 (x86 32-bit) with clang 1.1 and GNU libc 2.11.1;
- Ubuntu 12.04 (x86 64-bit) with gcc 4.6.3 and GNU libc 2.15;
- Ubuntu 12.04 (x86 64-bit) with clang 3.4 and GNU libc 2.15;
- Fedora 23.10 (x86 64-bit) with gcc 5.3.1 and GNU libc 2.22;
- Fedora 23.10 (x86 64-bit) with clang 3.7.0 and GNU libc 2.22;
- FreeBSD 10.3 (AMD 64-bit) with gcc 4.8.5;
- FreeBSD 10.3 (AMD 64-bit) with clang 3.4.1;


guilhermeferreira:
I also cross compiled to some other architectures, like ARM and MIPS.

icraggs:
I like the idea, but for the time being at least I want to allow the existing Makefile build to work, because that is what is used in the Eclipse Hudson automated builds for Linux and MacOS: https://hudson.eclipse.org/paho/.

I presume, that's as simple as leaving the existing Makefile in the repo as well?  Also, as we're coming up to a release of Paho in early June, I'd probably like to defer incorporating these changes until after then, as I would like to address as many previous issues as I can before then.

bootstrap
```diff
@@ -0,0 +1,37 @@
...
+#     http://www.eclipse.org/legal/epl-v10.html
+#  and the Eclipse Distribution License is available at 
+#    http://www.eclipse.org/org/documents/edl-v10.php.
+# 
+#  Contributors:
+#     Guilherme Maciel Ferreira - initial implementation
+#*******************************************************************************/
+
+# Run all tools in the right order
+autoreconf --install
```
JuergenKosel, line 19: Why do you need the other steps?
Just calling "autoreconf --install" should already include the other steps of this script.

.gitignore
```diff
@@ -1,3 +1,18 @@
 /dep/
 /build/
+/src/VersionInfo.h
+/src/MQTTVersion
+.deps/
+.libs/
+autom4te.cache/
+config.*
+configure
+configure.*
```
JuergenKosel, line 10: Do you really want to ignore configure.ac ?

test/Makefile.am
```diff
@@ -0,0 +1,182 @@
...
+#*******************************************************************************/
+
+###############################################################################
+# Binaries names
+###############################################################################
+
+bin_PROGRAMS  =
+
+if PAHO_BUILD_SYNC
+bin_PROGRAMS += test1
```
JuergenKosel, line 24: I would use "check_PROGRAMS" instead of "bin_PROGRAMS" here. Because you usually want to run these binaries during "make check". But you do not want to install them.

guilhermeferreira:
Hi Ian, thanks for the feedback. 

I was not aware that the Makefile could not be replaced. I'd removed it because Automake creates Makefiles, overwriting the one already in the repository. I'll keep the Makefile like you suggested, and postpone the pull request until June's release.

guilhermeferreira:
Juergen, thanks very much for the review. I'll incorporate your suggestions!

PR #141, opened 2016-05-31
CMake fixes on Microsoft Windows

guilhermeferreira:
This branch incorporates some fix to build on Windows. I resort to CMake GUI to generates Visual C++ build files. I successfully built Paho C on Visual Studio 2015.

Tested the CMake build on the following systems:
- Windows 7 SP1 (64-bit) with Visual Studio/C++ 14 (2015);
- Fedora 23.10 (64-bit) with gcc 5.3.1 and GNU libc 2.22;
- Fedora 23.10 (64-bit) with clang 3.7.0 and GNU libc 2.22;
- FreeBSD 10.3 (64-bit) with clang 3.4.1;
- Ubuntu 12.04 (64-bit) with gcc 4.6.3 with GNU libc 2.15;
- Ubuntu 12.04 (64-bit) with clang 3.4.0 with GNU libc 2.15;


tbeu:
These are important fixes for CMake for VS. Every Win user that pulls the current master (like I did) cannot build the Win libs out of the box. Please consider to merge! Thanks.

icraggs:
The Visual Studio build should still work (apart from the introduction of CMake removing the necessary version file - I'll have to fix that).  I intend to keep the Windows build working.

Issue #145, opened 2016-06-17
MQTTAsync_disconnect occasionally starts reconnection


rpendleton:
I've been debugging a problem where the connection onSuccess callback is being called during a disconnect. I believe I've tracked it down to the disconnect being detected as a failed connection, which then causes a reconnect attempt. This problem only happens every once in a while, but I can usually get it to happen about one in every ten disconnects.

Here's a detailed log of the connection: [paho.mqtt.c.log](https://github.com/eclipse/paho.mqtt.c/files/4500528/paho.mqtt.c.log)

As you can see, the disconnect function is called on lines [73-91](https://gist.github.com/rpendleton/08ade58d4bc46f9840f3749c68d76b5c#file-paho-mqtt-c-log-L73-91), the disconnect is sent around lines [176-186](https://gist.github.com/rpendleton/08ade58d4bc46f9840f3749c68d76b5c#file-paho-mqtt-c-log-L176-186), and then a failed connection is detected on lines [197-202](https://gist.github.com/rpendleton/08ade58d4bc46f9840f3749c68d76b5c#file-paho-mqtt-c-log-L197-202), triggering a reconnect. This new connection eventually leads to the execution of the connect onSuccess callback rather than the expected disconnect onSuccess.

rpendleton:
Upon further inspection, it appears that the disconnect is setting `connect_state = -2`, but then the `MQTTAsync_checkTimeouts` has the following logic (roughly):

```
if (connect_state != 0 && elapsed > timeout)
{
    // connection timed out
    continue;
}

if (connect_state == -2)
    MQTTAsync_checkDisconnect();
```

This means that disconnects that occur after the initial connection timeout period may be caught as connection timeouts. Am I correct in assuming that I can just reorder them as follows?

```
if (connect_state == -2)
{
    MQTTAsync_checkDisconnect();
}
else if (connect_state != 0 && elapsed > timeout)
{
    // connection timed out
    continue;
}
```

rpendleton:
(Sorry about all of the duplicate commits.. It took me a while to figure out how the Eclipse IP Validation worked.)

icraggs:
Thanks for working this out and contributing the fix.  I'm glad you figured out the Eclipse IP validation!   I've merged the fix into the develop branch.

icraggs:
I think this is fixed.  Closing.

PR #147, opened 2016-06-20
rearrange timeout checks, fixes issue #145

rpendleton:
I haven't noticed any side-effects of the included patch, but I'm also not too familiar with the internals of the library. A detailed log of the original issue is in #145.


icraggs:
Merged into the develop branch for inclusion in the next service release.

Issue #144, opened 2016-06-16
SSL flag in command is not set properly if using `serverURI` in client


prateekbansal:
If we are not passing in multiple `serverURIs`, but just passing a single `serverURI` using `client`, the ssl option on client is not set and so ssl seems to be broken.

Specifically, the processing on the serverURI at line src/MQTTAsync.c:1005-1013 should happen outside of `if` check:

```
if (command->command.details.conn.serverURIcount > 0) { ...
```

rpendleton:
Correct me if I'm wrong, but it looks like it's still being set if you use `MQTTAsync_create`. Look at lines 450-456 of master:

``` c
#if defined(OPENSSL)
    else if (strncmp(URI_SSL, serverURI, strlen(URI_SSL)) == 0)
    {
        serverURI += strlen(URI_SSL);
        m->ssl = 1;
    }
#endif
```

I have SSL working fine in one of my projects and I'm using a single server with `MQTTAsync_create`. There was one bug I encountered though, which was that the `CMakeLists.txt` file is missing the compiler flag that actually enables SSL. So everything would compile fine, but it would try resolving the hostname with `ssl://` in it. My pull request (#149) fixes that.

prateekbansal:
@rpendleton the code seems to have changed since my last pull request (around a couple of weeks back) which had that issue. Possibly the issue got resolved because I don't see that code anymore.

icraggs:
The SSL tests are working in both Travis and AppVeyor, so I think it's all good.  Closing.

PR #149, opened 2016-06-20
add missing OPENSSL define to CMakeLists.txt

rpendleton:
It looks like at some point, the cc flag for `OPENSSL=1` was lost. Because of this, OpenSSL versions of the MQTT library don't actually include SSL support. This caused some weird issues such as attempting to resolve hostnames with the `ssl://` prefix and SSL not actually being used, despite compiling without any issues.


prateekbansal:
@rpendleton the code seems to have changed since my last pull request (around a couple of weeks back) which had that issue. Possibly the issue got resolved because I don't see that code anymore.

JuergenKosel:
Hello,
the issue is still there. I just trapped in while building for windows target with cmake and visual studio 2010. (That's why I have to use cmake instead of the existing .sln files.)

rpendleton:
@JuergenKosel You're correct, as the problem is still there. When @prateekbansal said the issue was potentially resolved in later changes, he was referring to issue #144 rather than this one.

JuergenKosel:
For visual studio 2010 I need to define WIN32_LEAN_AND_MEAN, if the macro OPENSSL is defined.
Otherwise compilation fails in src/StackTrace.c due to inconsistant/duplicate type definitions.
Maybe this macro should also be added to the cmake file.

BeLearner:
problem still exists.Please fix it.I spent a whole afternoon to figure out why I couldn't connect with SSL.It turns out to be a problem of the repository.So Please fix it as soon as possible

rpendleton:
@BeLearner Did you still experience problems after applying the changes from this pull request?

BeLearner:
@rpendleton No,I haven't apply the changes from this pull request.Let me have a try.

BeLearner:
I build again with *COMPILE_DEFINITIONS "OPENSSL=1"*.It works now.Thanks @rpendleton .Please merge this pull request to master as soon as possilbe so that others won't get confused.

icraggs:
Thanks for fixing this.

PR #153, opened 2016-06-23
Call connect failure on socket error during connection (fixes #152)

### LINKED ISSUE #152: MQTTAsync_connect occasionally never calls connect.onFailure
Just as a forewarning, I'm using the patch from my pull request #147. I don't remember seeing this issue before, but I also just updated to the newest version of the library.

After calling MQTTAsync_connect with a connection that will fail (because of a hostname, port, connection, etc), I don't always get a connect failed callback. I'm not sure if it's relevant, but I seem to be getting the following error when it happens:

```
Error from MQTTAsync_cycle() - removing socket 8
```

It looks like the `MQTTAsync_receiveThread` calls `MQTTAsync_disconnect_internal` if the connection is already made, which will call the connection lost callback eventually. However, if the socket isn't already connected, it calls `MQTTAsync_closeOnly` which sets `connect_state = 0`. Then, in the `MQTTAsync_checkTimeouts` function, `connect_state == 0`, so a timeout or failed connection is never detected.

I'm currently using the following patch, but I have to admit that this one is quite a bit scarier than the one in #147, and I'd like some guidance on the issue:

``` patch
diff --git a/src/MQTTAsync.c b/src/MQTTAsync.c
index f475c27..4959e12 100755
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -1689,8 +1689,22 @@ thread_return_type WINAPI MQTTAsync_receiveThread(void* n)
                                MQTTAsync_disconnect_internal(m, 0);
                                MQTTAsync_lock_mutex(mqttasync_mutex);
                        }
-                       else /* calling disconnect_internal won't have any effect if we're already disconnected */
+                       else
+                       {
+                               /* calling disconnect_internal won't have any effect if we're already disconnected */
                                MQTTAsync_closeOnly(m->c);
+
+                               if (m->shouldBeConnected == 1 && m->connect.onFailure)
+                               {
+                                       MQTTAsync_failureData data;
+
+                                       data.token = 0;
+                                       data.code = MQTTASYNC_FAILURE;
+                                       data.message = "TCP socket error";
+                                       Log(TRACE_MIN, -1, "Calling connect failure for client %s", m->c->clientID);
+                                       (*(m->connect.onFailure))(m->connect.context, &data);
+                               }
+                       }
                }
                else
                {
```

rpendleton:
This is the patch I mentioned in #152. As I mentioned there, I'm a little more uncomfortable with it, but figured I'd make a pull request in case this actually is the proper solution. I just don't know enough about the internals to judge whether it's the right approach or not.

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -1689,8 +1689,22 @@ thread_return_type WINAPI MQTTAsync_receiveThread(void* n)
 				MQTTAsync_disconnect_internal(m, 0);
 				MQTTAsync_lock_mutex(mqttasync_mutex);
 			}
-			else /* calling disconnect_internal won't have any effect if we're already disconnected */
+			else
+			{
+				/* calling disconnect_internal won't have any effect if we're already disconnected */
 				MQTTAsync_closeOnly(m->c);
+
+				if (m->shouldBeConnected == 1 && m->connect.onFailure)
+				{
+					MQTTAsync_failureData data;
+
+					data.token = 0;
+					data.code = MQTTASYNC_FAILURE;
+					data.message = "TCP socket error";
+					Log(TRACE_MIN, -1, "Calling connect failure for client %s", m->c->clientID);
+					(*(m->connect.onFailure))(m->connect.context, &data);
+				}
+			}
 		}
 		else
 		{


rpendleton:
After using this for a few weeks, I have noticed some side-effects. It looks like the disconnect callback occasionally gets called more than once per connection. I'm assuming this is likely a threading issue, but I could be wrong since I haven't looked into it too much.  I figure I'll still leave this hear, since it's a partial solution to the problem, but it will need some work first.

icraggs:
I've added a fix for this under issue #190 to the develop branch.  Thanks.

PR #159, opened 2016-07-26
Improve password handling

dpfrey:
Support binary passwords by treating passwords as uint8_t arrays rather than null terminated strings.  This pull request does not include updates to the test or sample code.  I have tried the synchronous client and it seems to work.  I want to gauge whether there is interest in this change before I go through the work of updating the tests and samples.  This change also enforces that a password may not be present without a username.

### DIFF
--- a/src/Clients.h
+++ b/src/Clients.h
@@ -164,7 +164,8 @@ typedef struct
 {
 	char* clientID;					/**< the string id of the client */
 	const char* username;					/**< MQTT v3.1 user name */
-	const char* password;					/**< MQTT v3.1 password */
+	const uint8_t* password;				/**< MQTT v3.1 password */
+	uint16_t passwordLength;				/**< password length in bytes */
 	unsigned int cleansession : 1;	/**< MQTT clean session flag */
 	unsigned int connected : 1;		/**< whether it is currently connected */
 	unsigned int good : 1; 			/**< if we have an error on the socket we turn this off */
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -2167,10 +2167,15 @@ int MQTTAsync_connect(MQTTAsync handle, const MQTTAsync_connectOptions* options)
 			goto exit;
 		}
 	}
-	if ((options->username && !UTF8_validateString(options->username)) ||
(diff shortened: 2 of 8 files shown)


dpfrey:
FYI: I get a 500 error when I click the login link on https://projects.eclipse.org, so I was unable to sign the CLA.

dpfrey:
I have now signed the eclipse CLA.  What do I need to do to get the checks run again?

ralight:
Don't worry - the person who would merge the code could do a check outside of the github pull request.

icraggs:
While this looks good in itself, I think I need to support the old way of adding passwords at the same time, for backwards compatibility.

icraggs:
Thanks for the contribution.  I've added an update for this to the develop branch, including tests.  My update is a bit different - changing the external data structures in a backward compatible way.

dpfrey:
Sorry I didn't have a chance to update the PR, but thanks for implementing this change.  There is one thing that my pull request added that yours does not do which is requiring that a password may only be present if a username is also present.  See https://github.com/eclipse/paho.mqtt.c/pull/159/commits/2774efe925fca48474b1e49f48d637edd170c521#diff-866e259bd7df83e722c88fdb7bb9f9a8R2176

Perhaps that should have been submitted as a separate pull request.

icraggs:
The MQTT protocol allows username and password to be included independently - a password can be sent without a username.  I figured that although it seems like a feature that wouldn't be used, or useful, that the client library shouldn't disallow it.

dpfrey:
My change was based on this line from the spec:

> [MQTT-3.1.2-22] If the User Name Flag is set to 0, the Password Flag MUST be set to 0.

I interpret this as a requirement to check that the username is present if the password is present.  Is this incorrect?

icraggs:
Well that statement is pretty clear, for sure.  You got me there.

Interestingly, that constraint seems to have been removed from the latest MQTT v5 draft, and I thought that it wasn't in 3.1 (or 3.1.1 for that matter).    My feeling is still to leave it as it is, then the application is at the mercy of the server to decide.  Alternatively, if we do check, we probably ought to return an error (no password without username) so that it's clear what we are doing.

PR #165, opened 2016-08-02
Implemented connection timeouts for getaddrinfo() to prevent system hanging

CJxD:
Socket_new() will now honour the connection timeout on GNU systems (only GNU has an asynchronous version of getaddrinfo()).

Also moved `_GNU_SOURCE` to the compiler flags as it just would not work as a define in the top of `Socket.h`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,14 +128,14 @@ MQTTLIB_AS_TARGET = ${blddir}/lib${MQTTLIB_AS}.so.${VERSION}
 MQTTVERSION_TARGET = ${blddir}/MQTTVersion
 
 CCFLAGS_SO = -g -fPIC $(CFLAGS) -D_GNU_SOURCE -Os -Wall -fvisibility=hidden -I$(blddir_work)
-FLAGS_EXE = $(LDFLAGS) -I ${srcdir} ${START_GROUP} -lpthread -lanl ${END_GROUP} -L ${blddir}
-FLAGS_EXES = $(LDFLAGS) -I ${srcdir} ${START_GROUP} -lpthread -lanl -lssl -lcrypto ${END_GROUP} -L ${blddir}
+FLAGS_EXE = $(LDFLAGS) -I ${srcdir} ${START_GROUP} -lpthread ${GAI_LIB} ${END_GROUP} -L ${blddir}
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -127,15 +127,15 @@ MQTTLIB_A_TARGET = ${blddir}/lib${MQTTLIB_A}.so.${VERSION}
 MQTTLIB_AS_TARGET = ${blddir}/lib${MQTTLIB_AS}.so.${VERSION}
 MQTTVERSION_TARGET = ${blddir}/MQTTVersion
 
-CCFLAGS_SO = -g -fPIC $(CFLAGS) -Os -Wall -fvisibility=hidden -I$(blddir_work)
-FLAGS_EXE = $(LDFLAGS) -I ${srcdir} -lpthread -L ${blddir}
-FLAGS_EXES = $(LDFLAGS) -I ${srcdir} ${START_GROUP} -lpthread -lssl -lcrypto ${END_GROUP} -L ${blddir}
+CCFLAGS_SO = -g -fPIC $(CFLAGS) -D_GNU_SOURCE -Os -Wall -fvisibility=hidden -I$(blddir_work)
```
guilhermeferreira, line 130: Once you removed the _GNU_SOURCE from the source files, I guess that  _GNU_SOURCE should be in CMakeLists.txt too. Or maybe in a common file, so both build systems can use. Otherwise the CMake build will differ from the build performed by Make.

icraggs:
There are several problems with this PR as it stands as shown in the following build:

https://travis-ci.org/eclipse/paho.mqtt.c/jobs/225162289

icraggs:
I know it's a long time since this PR was created, but I've had to comment out the code as it intermittently gets a -103 ECONNABORTED in the CI tests.  I don't know why yet - it does not seem to be a simple timeout.  If anyone has any ideas I'd like to hear them.

Makefile
martinboers: In addition to adding the _GNU_SOURCE  compile definition to CMakeLists.txt, I believe that `anl` (asynchronous name lookup) must also be added to the set of LIBS_SYSTEM libraries.

PR #175, opened 2016-09-11
Protect SSL code from empty strings

guilhermeferreira:
The SSLSocket_createContext() uses MQTTClient_SSLOptions's members
(keyStore, trustStore and enabledCipherSuites) even if they are empty
strings.

Add empty string, besides NULL pointer, as an invalid value. Because
the functions SSL_CTX_use_certificate_chain_file(), SSL_CTX_load_verify_locations()
and SSL_CTX_set_cipher_list() also issue error if the arguments are
empty strings.

### DIFF
--- a/src/SSLSocket.c
+++ b/src/SSLSocket.c
@@ -474,8 +474,8 @@ int SSLSocket_createContext(networkHandles* net, MQTTClient_SSLOptions* opts)
 			SSLSocket_error("SSL_CTX_new", NULL, net->socket, rc);
 			goto exit;
 		}
-	
-	if (opts->keyStore)
+
+	if (opts->keyStore && (strlen(opts->keyStore) > 0))
 	{
 		int rc1 = 0;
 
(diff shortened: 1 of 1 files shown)


src/SSLSocket.c
```diff
@@ -552,11 +552,12 @@ int SSLSocket_setSocketForSSL(networkHandles* net, MQTTClient_SSLOptions* opts)
 	if (net->ctx != NULL || (rc = SSLSocket_createContext(net, opts)) == 1)
 	{
 		int i;
+
 		SSL_CTX_set_info_callback(net->ctx, SSL_CTX_info_callback);
```
JuergenKosel, line 556: In this line (and later) a NULL pointer access could occur, because net->ctx could be NULL here!
This is going to be fixed in #161

JuergenKosel: The return code is needed this way because MQTTClient_connectURIVersion() and MQTTAsync_connecting() expect this.
One could change also this functions. But then this bug fix will become a never ending story.

icraggs:
I've fixed the return code from createContext in the develop branch.

Do we really need these extra checks too?  There are presumably many other ways that the strings could be invalid, leading to a non-existent filename.  As long as the OpenSSL functions don't return 1 (for success), then we will be ok.   I ran a test with all the strings set to "", and it worked (or failed) as expected (after the createContext fix).

icraggs:
I'm assuming that the thumbs up means agreement!  I think that the createContext fix covers all these cases, so I'll close this PR.  Reopen if you can show otherwise.

Thanks for the contribution.

guilhermeferreira:
Hello @icraggs. Yes, let it closed. I'm going to reopen if, after my tests, I find out we still need this.

PR #178, opened 2016-09-15
Automated coverity scan builds.

ralight:
This addition to the travis configuration allows us to easily submit builds to the static analysis tool coverity scan.

The configuration means that only branches with the pattern "coverity-._" get sent through to coverity, so the procedure is to carry on development in the master and other branches, then merge to a "coverity-._" branch when the scan is wanted. I'd suggest creating a coverity-master and coverity-develop branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,5 +8,18 @@ os:
   - linux
   - osx
   
+env:
+  global:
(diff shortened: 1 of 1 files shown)


icraggs:
Thanks Roger.  I've merged this into the develop branch.

Issue #179, opened 2016-09-22
Second call of MQTTAsync_connect() fails always for IPv6 server URI


JuergenKosel:
Hello,

if an IPv6 server URI without port number is used, than MQTTAsync_connect() could be used only once:
1. Call MQTTAsync_create()
   Output from gdb:
   `
   Breakpoint 1, MQTTAsync_create (handle=0x7179a02c, serverURI=0x717a6464 "tcp://[2002:3e92:37e2:0:250:56ff:fe28:1055]", 
     clientId=0x7179a030 "00:06:71:33:00:13", persistence_type=1, persistence_context=0x0)
     at paho.mqtt.c/src/MQTTAsync.c:506
   `
2. Then first call of  MQTTAsync_connect()
   Output from gdb:
   `
   Breakpoint 2, MQTTAsync_connect (handle=0x717d6e2c, options=0x717d7550) at paho.mqtt.c/src/MQTTAsync.c:2141
   2141    in paho.mqtt.c/src/MQTTAsync.c
   (gdb) next
   2142    in paho.mqtt.c/src/MQTTAsync.c
   (gdb) print *m
   $2 = {serverURI = 0x717d6f8c "[2002:3e92:37e2:0:250:56ff:fe28:1055]", ssl = 0, c = 0x717d7184, 
   cl = 0x5de01 <MqttBroker::MQTTAsync_connectionLost(void*, char*)>, 
   ma = 0x5dd99 <MqttBroker::MQTTAsync_messageArrived(void*, char*, int, MQTTAsync_message*)>, 
   dc = 0x5dda1 <MqttBroker::MQTTAsync_deliveryComplete(void*, int)>, context = 0x71799fe8, 
   `

With this call the connection to the broker could be established.
1. Restart the broker, so that the current connection will be lost and a new one could be established (if there is not this issue).
2. Another call of MQTTAsync_connect() should establish the connection:
   `
   Breakpoint 2, MQTTAsync_connect (handle=0x717d6e2c, options=0x717d7550) at paho.mqtt.c/src/MQTTAsync.c:2141
   2141    in paho.mqtt.c/src/MQTTAsync.c
   (gdb) next
   2142    in paho.mqtt.c/src/MQTTAsync.c
   (gdb) print *m
   $3 = {serverURI = 0x717d6f8c "[2002:3e92:37e2:0:250:56ff:fe28:1055", ssl = 0, c = 0x717d7184,
   `

[...]

PR #180, opened 2016-09-22
Fix usage of serverURIs to prevent segmentation fault

JuergenKosel:
The member serverURIs of type MQTTAsync_disconnectOptions was not usable in C++, because it was declared as read-only type: char\* const*
After this change, the server URI can now be provided within
MQTTAsync_disconnectOptions. So that issues as described in
https://github.com/eclipse/paho.mqtt.c/issues/179 could be avoided.

Signed-off-by: Juergen Kosel juergen.kosel@softing.com

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -1824,7 +1824,8 @@ static thread_return_type WINAPI MQTTAsync_receiveThread(void* n)
 							MQTTAsync_successData data;
 							memset(&data, '\0', sizeof(data));
 							Log(TRACE_MIN, -1, "Calling connect success for client %s", m->c->clientID);
-							if (m->serverURIcount > 0)
+							if ((m->serverURIcount > 0)
+							    && (m->connect.details.conn.currentURI < m->serverURIcount))
 								data.alt.connect.serverURI = m->serverURIs[m->connect.details.conn.currentURI];
 							else
 								data.alt.connect.serverURI = m->serverURI;
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -1168,17 +1168,31 @@ static int MQTTAsync_processCommand(void)
 
 			if (command->client->serverURIcount > 0)
 			{
-				serverURI = command->client->serverURIs[command->command.details.conn.currentURI];
-
-				if (strncmp(URI_TCP, serverURI, strlen(URI_TCP)) == 0)
-					serverURI += strlen(URI_TCP);
-#if defined(OPENSSL)
-				else if (strncmp(URI_SSL, serverURI, strlen(URI_SSL)) == 0)
+				if (command->client->c->MQTTVersion == MQTTVERSION_DEFAULT)
 				{
-					serverURI += strlen(URI_SSL);
-					command->client->ssl = 1;
+					if (command->command.details.conn.MQTTVersion == MQTTVERSION_3_1)
+					{
+						command->command.details.conn.currentURI++;
+						command->command.details.conn.MQTTVersion = MQTTVERSION_DEFAULT;
+					}
 				}
+				else
+					command->command.details.conn.currentURI++;
+
+				if (command->command.details.conn.currentURI < command->client->serverURIcount)
+				{
+					serverURI = command->client->serverURIs[command->command.details.conn.currentURI];
+
+					if (strncmp(URI_TCP, serverURI, strlen(URI_TCP)) == 0)
+						serverURI += strlen(URI_TCP);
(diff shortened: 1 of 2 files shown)


icraggs:
@fpagliughi any comments on this one?

fpagliughi:
I just started writing a really long comment, then realized I was talking about the wrong thing (too much const, instead of not enough)!

I agree with Juergen, but over an even wider range. I don't thing that the C library should alter the contents _any_ buffer sent to it (string or binary)... and it should properly promise that it won't. So string variables should always be:
```
const char* userName;
```
and binary buffers should preferably be:
```
const void* payload;
size_t payloadlen;
```
The C lib has been evolving towards this for the last few years which makes me very happy.

(BTW, using `size_t` instead of `int` for array lengths often helps in portability between 32 and 64-bit systems, and helps with C++ compatibility).

Using const is helpful if you want to use quoted string literals. In C++ these are of type `const char*`, and in either language, you wouldn't want to modify the contents!

So if the library wants to alter a string, it should make a local copy into variable memory, ensure that it has enough space, and then alter the local copy.

But I would take Jurgen's modification of `serverURIs` one step further:
```
const char* const* serverURIs;
```
(if I have this correct).
The variable/field `serverURIs` should remain variable; it can be changed by the application later to point to something else. But the library should promise not to change **anything** that the variable points to; not the URI strings themselves, and it should also not alter the array of pointers to those strings. IMHO.

JuergenKosel:
Hello,
the problem with the char** const declaration was, that the C++ application, which wants to use the library can not set the serverURIs.
Maybe we need to add a C++ unit test here. At least to see, that it compiles within C++.

Greetings
Juergen

fpagliughi:
Hello Jurgen,

The extra const would create a trade-off. It would say that anything using the struct directly could not change the pointers to the strings from the structure. Even the application. But that is a worthwhile trade-off since, really it means that the library is promising that it won't change the contents of the array of pointers that the application hands it.

The application still has the option of setting up the array in advance and then assigning it to the structure by assigning the pointer. The app could then use an array of literals or build up an array dynamically, and assign it to the struct.

This all works in C++:

```
struct Options {
    const char* const* serverURIs;
};

int main()
{
	const char* const uris[] = {
		"tcp://server1:1883",
		"tcp://server2:1883",
		"ssl://server3:8883"
	};

	Options opts { uris };

	// or...
	opts.serverURIs = uris;

	const int N = 5;
	const char **dynamicUris = new const char*[N];
	for (int i=0; i<N; ++i)
		dynamicUris[i] = new char[20];

	opts.serverURIs = dynamicUris;
    //...
}
```

The only thing I couldn't figure out how to do was a totally static initialization for an Options object.

fpagliughi:
I would propose we do the same thing with subscribeMany and unsubscribeMany - the collection of topic names should be this same type of structure with the same guaranteed of const-ness. 

```
int MQTTAsync_subscribeMany(MQTTAsync handle, int count, const char* const* topics, 
    int* qos, MQTTAsync_responseOptions* response);

int MQTTAsync_unsubscribeMany(MQTTAsync handle, int count, const char* const* topics,
    MQTTAsync_responseOptions* response);
```

Perhaps I should create a new issue (or pull request) for this one?

fpagliughi:
And not to belabor the point, but either way would still be compatible with a contiguous SLT container, like a vector:
```
    std::vector<const char*> svec;
    svec.push_back("tcp://server1:1883");
    svec.push_back("tcp://server2:1883");
    svec.push_back("ssl://server3:8883");

    opts.serverURIs = svec.data();
```

With the extra const in the serverURI, this works with both const and non-const svec.

JuergenKosel:
Hello @fpagliughi
Please create another pull request, to put not to many in this one.

icraggs:
Now this proposed change is causing compile warnings on the Linux/Mac builds.  

I conducted an exercise about 18 months ago where I analyzed the const changes that could be made without affecting existing applications.  From my perspective, it's a pain to cause a C application to add const modifiers just to get a clean compile.  The current approach added as much instances of const as we could without forcing extra effort on the application programmer.  

I should have pointed this out earlier in the PR, but I thought from the title that this was about relaxing the conditions, not tightening them.

JuergenKosel:
Yes, the pull request was originally to relax the conditions.
But beside the modified declaration in the header file, you should take a look at the other modifications.
Because I had run into several segmentation faults by using the serverURIs array.
So please cherry pick the commits which avoid the access of serverURIs[i] where i is >= the number of URIs in the array.

JuergenKosel:
#539 replaces this PR

PR #181, opened 2016-10-08
Build-system refactoring

rpoisel:
- CMake function calls renamed to match function names from the CMake documentation
- the OpenSSL library will now be searched for by issuing the official `find_package()` command
- using the `target_include_directories()` and `target_link_libraries()` commands instead of general `include_directories()` and `link_directories()` commands to simplify future maintenance work
- Version info is now kept in only "one" place for all (CMake, GNU Make, Apache Ant) build systems: `version.major`, `version.minor` and `version.patch` files in the root directory of this repo
- toolchain files for the `linux-x64`, `linux-x86` platforms
- implicit function declarations will now be treated as errors
- rename of `#include <Windows.h>` to `#include <windows.h>` (as mentioned in MSDN); this is required for cross-compilation from case-sensitive systems such as Linux to Windows platforms


rpoisel:
Please have a look at the results of your CI. I do not understand, why the build fails. The only thing that has changed is the way the version string is determined. Thx!

src/CMakeLists.txt
```diff
@@ -45,53 +40,65 @@ SET(common_src
...
-    DESTINATION include)
+    LIBRARY DESTINATION lib
+    ARCHIVE DESTINATION lib
+)
+install(FILES MQTTAsync.h MQTTClient.h MQTTClientPersistence.h
+    DESTINATION include
+)
 
-IF (PAHO_WITH_SSL)
-    SET(OPENSSL_LIB_SEARCH_PATH "" CACHE PATH "Directory containing OpenSSL libraries")
```
guilhermeferreira, line 73: @rpoisel, I'm not sure if Travis uses CMake to build. But you've removed the OpenSSL paths and the OSX build fails because it doesn't find OpenSSL headers. This is the problem or just a coincidence.

```
In file included from src/Clients.c:24:
src/Clients.h:27:10: fatal error: 'openssl/ssl.h' file not found
#include <openssl/ssl.h>
         ^
```

rpoisel: @guilhermeferreira, thanks for having a look, but that does not seem to be the case. See [here](https://travis-ci.org/eclipse/paho.mqtt.c/jobs/166283011), line 47. Travis CI is using plain old GNU make to build paho. I guess that openssl needs to be installed, e.g. by using `brew` or so. Or, maybe the build node has been reinstalled without openssl?

icraggs:
Hello all.  Thanks for the contributions.  As I'm responsible for several Paho packages I'm finding it difficult to fit in the time to look at all the build updates for the C client.  I will try to do so soon.

guilhermeferreira:
@rpoisel, how did you solve the build error?

CMakeLists.txt
```diff
@@ -16,40 +16,45 @@
...
+project("paho" C)
+cmake_minimum_required(VERSION 3.0)
+
+if (NOT CMAKE_TOOLCHAIN_FILE)
+    message(FATAL_ERROR
+        "Paho has to be built using a toolchain file. \
+These can be found in ${CMAKE_CURRENT_LIST_DIR}/cmake and can be \
+specified using the CMAKE_TOOLCHAIN_FILE variable when invoking cmake."
+    )
+endif ()
```
LeSpocky, line 29: Why requiring a toolchain file? What about native builds?

CMakeLists.txt
```diff
@@ -16,40 +16,45 @@
...
-STRING(STRIP ${BUILD_TIMESTAMP} BUILD_TIMESTAMP)
+string(TIMESTAMP BUILD_TIMESTAMP UTC)
 
 ## build options
-SET(PAHO_WITH_SSL FALSE CACHE BOOL "Flag that defines whether to build ssl-enabled binaries too. ")
-SET(PAHO_BUILD_DOCUMENTATION FALSE CACHE BOOL "Create and install the HTML based API documentation (requires Doxygen)")
-SET(PAHO_BUILD_SAMPLES FALSE CACHE BOOL "Build sample programs")
+set(PAHO_WITH_SSL FALSE CACHE BOOL "Flag that defines whether to build ssl-enabled binaries too. ")
+set(PAHO_BUILD_DOCUMENTATION FALSE CACHE BOOL "Create and install the HTML based API documentation (requires Doxygen)")
+set(PAHO_BUILD_SAMPLES FALSE CACHE BOOL "Build sample programs")
```
LeSpocky, line 42: You could use the 'option' command to have those build options readily available and visible e.g. in CMake gui on first build. Those can still be overwritten from an outside build system like buildroot or ptxdist.

cmake/toolchain.linux-x64.cmake
```diff
@@ -0,0 +1,17 @@
...
+set(CMAKE_SYSTEM_VERSION 1)
+
+# specify the cross compiler
+set(CMAKE_C_COMPILER gcc)
+set(CMAKE_SHARED_LINKER_FLAGS "-Wl,--no-undefined" CACHE STRING "" FORCE)
+set(CMAKE_EXE_LINKER_FLAGS "-Wl,--no-undefined" CACHE STRING "" FORCE)
+
+# compiler flags
+set(CMAKE_C_FLAGS "--std=gnu99 -Werror=implicit-function-declaration" CACHE STRING "" FORCE)
+
```
LeSpocky, line 14: Those --std options may be set with the CMake variables CMAKE_C_STANDARD and CMAKE_C_STANDARD_REQUIRED in the main CMake files.

btw: I see no sense in providing toolchain files at all, those are highly user and toolchain specific. Cross build systems come with their own toolchain files anyway.

src/CMakeLists.txt
```diff
@@ -45,53 +40,71 @@ SET(common_src
...
-ADD_LIBRARY(paho-mqtt3c SHARED ${common_src} MQTTClient.c)
-ADD_LIBRARY(paho-mqtt3a SHARED ${common_src} MQTTAsync.c)
-TARGET_LINK_LIBRARIES(paho-mqtt3c pthread ${LIBS_SYSTEM}) 
-TARGET_LINK_LIBRARIES(paho-mqtt3a pthread ${LIBS_SYSTEM})
-TARGET_LINK_LIBRARIES(MQTTVersion paho-mqtt3a paho-mqtt3c ${LIBS_SYSTEM})
-SET_TARGET_PROPERTIES(
+add_executable(MQTTVersion MQTTVersion.c)
+add_library(paho-mqtt3c SHARED ${common_src} MQTTClient.c)
+add_library(paho-mqtt3a SHARED ${common_src} MQTTAsync.c)
+target_include_directories(paho-mqtt3c PUBLIC
```
LeSpocky, line 54: You might want to remove those "SHARED" options and use the global BUILD_SHARED_LIBS mechanism, maybe also as an option to set it to shared by default. This way the user can decide if she wants to build static or shared.

LeSpocky (commented):
In general I very much appreciate better CMake support for this lib. In detail some things don't seem to be made the CMake way. Additionally it would not hurt to split this whole stuff into several commits for easier review, especially just changing case of CMake command invocations could be in a separate commit.

andnig:
Any updates on this? The case sensitivity fix for windows.h would be awesome. And thanks for your efforts!

icraggs:
I'm not an expert on CMake, so I don't understand the rationale for some of the changes.  { to ( for instance.   What is the case sensitivity fix for windows.h, and how does it help?  The builds are working at the moment on Linux, Mac and Windows.

andnig:
Maybe I mistake the issue, however I'm cross-compiling the library on ubuntu as stated in the Readme-section for Windows 64/32.
For Version 1.2.1, this works perfectly fine, however the latest master commits fail, because my case sensitive cross-compiler provides an include "windows." but not a "Windows.h". As I was able to recognize, the source code sometimes includes windows.h and sometimes Windows.h (which is perfectly fine, as long as you build on windows (as this world is not case sensitive))

icraggs (changes requested):
I realize I've put off looking at this for a long time, it was the extent of the changes that was daunting.
I don't know why the case of keywords in the CMakeLists files should matter, and those changes swamp the other useful ones.  It would have been better structured into 3 at least I think 1) CMakeLists case changes (+ justification) 2) Makefile {} to () changes and 3) the other substantial ones like the version number in one place (v useful).  

I'm going to try to extract the substantial changes in 3)

waltronix:
@icraggs I can support with extracting the version stuff to reduce the changeset. Would that help?

icraggs:
@waltronix yes, if you could extract the version number consolidation support into one PR, that would be a great help

icraggs:
Single source files for version nos now merged in #796.  Windows header file names references now lower case.  Thanks.

PR #189, opened 2016-11-14
allow user to tell library to Skip openssl init

loigu:
Hi,

I'm using paho mqtt in application alongside with other libraries that also uses openssl. The thing is, that paho MQTTClient always initialize openssl in _create and deinitialize on _destroy. Which causes havoc for following reasons:

    on calling MQTTClient_create I have openssl library already initalized by external code and paho does the initialization again in time when openssl is already in use
    on calling MQTTClient_create paho overrides the already initialized openssl locking (duplicating the locks)
    on calling MQTTClient_destroy it deinitalizes openssl library that is still in use by other parts of my app

what I propose is to allow user of paho mqtt to set globally who is responsible for openssl library init.

I would appreciate any comments and hope for possible merge to upstream

Thanks

Jirka

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -1322,8 +1322,11 @@ void MQTTAsync_checkTimeouts()
 		
 		MQTTAsyncs* m = (MQTTAsyncs*)(current->content);
 		
+		/* check disconnect timeout */
+		if (m->c->connect_state == -2)
+			MQTTAsync_checkDisconnect(m, &m->disconnect);
 		/* check connect timeout */
-		if (m->c->connect_state != 0 && MQTTAsync_elapsed(m->connect.start_time) > (m->connectTimeout * 1000))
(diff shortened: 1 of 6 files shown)


loigu:
the build on mac fails due to code from develop branch, not affected by the change.

JuergenKosel:
@loigu in pull request #192 the continous integration build fails also due to missing openssl/ssl.h
I guess this file has been deleted by accident. :-(

icraggs:
Merged into develop, thanks.  I am wondering if the global init function should take a structure rather than one value, just in case some other future settings come along.

loigu:
Using structure (and init function for the structure) will make adding new features easier and backward compatible for sure.
I was thinking about it when making the patch but in the end decided for this simple straightforward solution because I don't see into paho.mqtt much. If you think there might be new options coming along in the future then using flags or structure is definitely better solution.

icraggs:
Thanks.   I've integrated the changes, and may well change the init parm to a structure.

PR #202, opened 2016-12-28
Support paho.mqtt.c build on Android.

leeminghao:


icraggs:
To allow me to accept this contribution, you would need to sign the eclipse CLA.  Also the Makefiles contain Android copyright and license statements - the easiest approach is to remove them, which I think in this case should be fine as they contain minimal Android related information.

icraggs:
Replaced by PR #299, as the IP checks failed on this one.

PR #213, opened 2017-01-13
RPM Specs for Fedora, RHEL and Centos

orpiske:
This pull request contains the required changes for packaging Paho C as a RPM. Building the RPM spec as provided in this PR results in 3 separate RPM packages: 
- paho-c 1.1.0: runtime
- paho-c-devel 1.1.0: development files (headers and examples)
- paho-c-devel-docs 1.1.0: development documentation.

I currently have these packages on a testing COPR for Fedora 23 to 26, as well as EL7: https://copr.fedorainfracloud.org/coprs/orpiske/paho-testing/.

Please let me know if you would like any change to be made to this.

### DIFF
--- a/dist/Makefile
+++ b/dist/Makefile
@@ -1,0 +1,8 @@
+VERSION=1.1.0
+
+rpm-prep:
+	mkdir -p ${HOME}/rpmbuild/SOURCES/
+	tar --transform="s/\./paho-c-${VERSION}/" -cf ${HOME}/rpmbuild/SOURCES/paho-c-${VERSION}.tar.gz --exclude=./build --exclude=.git --exclude=*.bz ./ --gzip
+
+rpm: rpm-prep
+	rpmbuild -ba dist/paho-c.spec
--- a/dist/paho-c.spec
+++ b/dist/paho-c.spec
@@ -1,0 +1,65 @@
+%global _enable_debug_package 0
+%global debug_package %{nil}
+
+Summary:            MQTT C Client
+Name:               paho-c
+Version:            1.1.0
+Release:            1%{?dist}
+License:            Eclipse Distribution License 1.0 and Eclipse Public License 1.0
+Group:              Development/Tools
+Source:             paho-c-%{version}.tar.gz
+URL:                https://eclipse.org/paho/clients/c/
+BuildRequires:      cmake
+BuildRequires:      gcc
+BuildRequires:      graphviz
+BuildRequires:      doxygen
+BuildRequires:      openssl-devel
+Requires:           openssl
+
+
+%description
+The Paho MQTT C Client is a fully fledged MQTT client written in ANSI standard C.
+
+
+%package devel
+Summary:            MQTT C Client development kit
+Group:              Development/Libraries
+Requires:           paho-c
+
+%description devel
+Development files and samples for the the Paho MQTT C Client.
+
+
+%package devel-docs
+Summary:            MQTT C Client development kit documentation
+Group:              Development/Libraries
+
+%description devel-docs
+Development documentation files for the the Paho MQTT C Client.
+
+%prep
+%autosetup -n paho-c-%{version}
+
+%build
+mkdir build.paho && cd build.paho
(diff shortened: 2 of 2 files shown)


orpiske:
I wonder why it keeps saying the validation failed, even though I signed it off w/ my correct ID.

icraggs:
Could really do with the last commit IP validation being resolved.  Could you try resubmitting?  As it's such a small change, an option would be to just remove it, and I could add it.

orpiske:
@icraggs Absolutely. I will try to rebase this, leaving the last commit out. Hopefully that will be sufficient for solving this issue. If that doesn't work, I'll open a separate PR.

orpiske:
@icraggs Rebasing the whole PR seems to have solved the problem. I had opened a separate PR without the offending commit, but I will close that one since this already contains it.

icraggs (approved):
I haven't tried it, but as it doesn't affect anything else, it's fine.

icraggs:
Do you have some doc on how to use?

icraggs:
I've merged the changes into the develop branch.

orpiske:
@icraggs I don't have any documentation but I can contribute one. The basic operation, from a RPM packaging point of view, is pretty simple. I have made it in such a way that the user/developer can choose between: 

1. Self packaging (ie.: whoever downloads/builds Paho, can create their own packages):
In this case, the developer can run: make -f dist/Makefile rpm and that would generate the RPM file. It's also totally possible to generate the RPM file manually, but just running "rpmbuild -ba dist/paho-c.spec" (provided the initial RPM build setup is in place). 

2. Generate the RPMs via Fedora COPR: in this case, the package is built by Fedora packaging infrastructure and a set of repos for RHEL, CentOS and multiple Fedora versions are setup automatically. The steps are not much more complicated that the previous one: 

2.1 Choose a MockSCM project type.
2.2 Fill in with the following details:

> Source Type: Mock build from a SCM repository
> SCM type: git
> SCM URL: https://github.com/eclipse/paho.mqtt.c
> SCM branch: <empty>
> Path to .spec file: dist/paho-c.spec

A sample project can be found here: https://copr.fedorainfracloud.org/coprs/orpiske/paho-testing/ (it can cloned and modified to point to Eclipse github repo instead of mine).

icraggs:
When I run "rpmbuild -ba dist/paho-c.spec" I get:

+ cmake -DPAHO_WITH_SSL=TRUE -DPAHO_BUILD_DOCUMENTATION=TRUE -DPAHO_BUILD_SAMPLES=TRUE -DCMAKE_INSTALL_PREFIX=/home/icraggs/rpmbuild/BUILDROOT/paho-c-1.1.0-1.el6.x86_64/usr ..
CMake Error: The source directory "/home/icraggs/rpmbuild/BUILD/paho-c-1.1.0" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
error: Bad exit status from /var/tmp/rpm-tmp.iktA3w (%build)


RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.iktA3w (%build)

orpiske:
@icraggs that seems to be caused because the tarball containing the source code contains the build.paho directory. 

I managed to reproduce the problem with: 
`mkdir build.paho ; export VERSION=1.1.0 ; tar --transform="s/\./paho-c-${VERSION}/" -cf ${HOME}/rpmbuild/SOURCES/paho-c-${VERSION}.tar.gz --exclude=./build --exclude=.git --exclude=*.bz ./ --gzip && rpmbuild -ba dist/paho-c.spec`


Whereas: 

1. Cleanup
`rm -rf build.paho && rm -rf ~/rpmbuild/SOURCES/paho* `

2. Repackage without build.paho: 
`export VERSION=1.1.0 ; tar --transform="s/\./paho-c-${VERSION}/" -cf ${HOME}/rpmbuild/SOURCES/paho-c-${VERSION}.tar.gz --exclude=./build* --exclude=.git --exclude=*.bz ./ --gzip && rpmbuild -ba dist/paho-c.spec`

Results in a successful build. Looking at the code, it looks like the package Makefile contains a reference to only the build dir (and not build.paho). I can send a PR for that as well.

orpiske:
@icraggs I have just opened PR #274 to ensure that the build.paho directory is excluded from the RPM source tarball generation. Users relying on the RPM Makefile to create the tarball won't have this problem.

PR #214, opened 2017-01-17
Proposed fix for OpenSSL compilation issue on OS X

orpiske:
This is a proposed fix for the compilation issue on newer versions of OS X. This has made the compilation fail on OS X - thus causing the build errors seem on Travis CI. The patch modifies the build so that it uses OpenSSL files provided by Brew. It also switches to FIND_PATH/FIND_LIBRARY functions provided by cmake. 

A sample build result is available here: https://travis-ci.org/orpiske/paho.mqtt.c/builds/192594544.

PR #222, opened 2017-02-07
Add support for building tests with CMake

orpiske:
This is the first PR on a set of PRs to follow and it adds support for building the tests using the CMake build system. On this patch, there is no elaborate logic to turn on/off tests based on the OS being used nor the build features. The subsequent PRs, which should improve testability and Travis/AppVeyor CI support for Linux and Windows will contain an adequate logic to handle that.

The code in this PR was tested on OS X, Linux (Fedora, RHEL) and FreeBSD. 

This PR is dependent on/related to [PR 214](https://github.com/eclipse/paho.mqtt.c/pull/214), which contains a proposed fix for OpenSSL compilation issues on OS X.


icraggs:
Merged into the develop branch.  Thanks

PR #225, opened 2017-02-14
Add support for test case execution on Travis

orpiske:
This is a follow up PR on [PR 222 ](https://github.com/eclipse/paho.mqtt.c/pull/222) and it adds support for building and running the unit tests on Travis CI.

This is still a work in progress and some work is needed to cleanup the build code, but it should give an idea of the PR intentions.

The code tries to leave the orchestration part of the test to the CI environment. That makes it possible use different MQTT brokers as needed by the users.


icraggs:
Thanks.  I've merged these changes into the develop branch, and made a few changes on top to get the tests to work more reliably.  There were a couple of defects that I found and fixed as a result.

PR #236, opened 2017-03-19
Documentation updates

orpiske:
A small patch that include updated test details and commands to the README file.


icraggs:
Thanks

PR #241, opened 2017-04-05
Cmake build dir cleanup

orpiske:
Ian, these are two very small changes I would like to propose: 
1. Removes specific test scripts from the build directory, since they are now being managed by cmake itself. 
2. Add the build.paho directory to the gitignore, to prevent it from being incorrectly added to the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /dep/
 /build/
+/build.paho/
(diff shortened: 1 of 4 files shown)


icraggs:
Thanks.  I've merged the changes into the develop branch.

PR #192, opened 2016-11-17
Issue 190

JuergenKosel:
This pull request provides a fix for #190 

Commit 0bd845fec90fc1f2d77e74757f2e50db4f42bd4c fixes a race condition.
After applying this commit, issue 190 occured each time. Which allowed me to provide a fix for #190 in the 2 following commit.

PR #249, opened 2017-04-10
Regression check vs uncover issue

JuergenKosel:
This is a test/verification pull request to verify if #192 introduces a new regression issue or uncovers an already existing issue.


icraggs:
This looks good.  I've merged it into develop.

PR #250, opened 2017-04-12
message token

lun0144:
change some code for message token can use for message ID.


icraggs:
I don't think this is correct.  It causes tests to fail.  The message id token is returned to the application in the response structure.

lun0144:
But according to my test，If do not modify it,the return message token will not be correct.

icraggs:
The message id is assigned by the client library, and returned to the application.  It's not up to the application to assign it.

lun0144:
ok, thank you .

PR #278, opened 2017-05-13
Readme: properly document OPENSSL_SEARCH_PATH

Bubu:
Since 28a3114ce97a5e the documentation regarding openssl search paths we pass
to cmake is out of date.

Properly document the current behaviour.

*Note:* Please consider merging this into master asap because that is the most visible build documentation and it's wrong there too.


icraggs:
Thanks.

PR #290, opened 2017-06-23
Fix `serverURIs` iterating in MQTTAsync.c

lancechentw:
This commit fixes two bugs related to `serverURIs` iterating in MQTTAsync.c

1. Segfault occurs when there is no healthy connection in `serverURIs`

   Two off-by-one errors of the `currentURI` cause the issue. The first
   one happens when doing boundary check of `serverURIs`, and the other
   one happens when accessing `serverURIs` without checking the boundary
   first.

2. The first URI in `serverURIs` gets skipped when restricting
   `MQTTVersion` to a particular value

   `currentURI` gets increased *before* connection attempts in each
   iteration.

Signed-off-by: Lance Chen <cyen0312@gmail.com>

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -430,7 +430,7 @@ static int MQTTAsync_checkConn(MQTTAsync_command* command, MQTTAsyncs* client)
 	int rc;
 
 	FUNC_ENTRY;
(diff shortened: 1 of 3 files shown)


icraggs:
Thanks

PR #295, opened 2017-07-04
Support for VxWorks 6.9

rpoisel:
These are the modifications we did to make Paho working on VxWorks 6.9. Actually, I'd prefer to create an operating system layer without all these #if and #else directives, but this would be a much more fundamental commit.

All in all, modifications where not severe. C++ comments were modified to ANSI-C style comments and a few operating system functions had to be replaced or customized.

### DIFF
--- a/src/Heap.c
+++ b/src/Heap.c
@@ -35,7 +35,6 @@
 char* Broker_recordFFDC(char* symptoms);
 #endif /* HEAP_UNIT_TESTS */
 
-#include <memory.h>
 #include <stdlib.h>
 #include <string.h>
 #include <stdio.h>
@@ -228,7 +227,7 @@ static int Internal_heap_unlink(char* file, int line, void* p)
 		Log(TRACE_MAX, -1, "Freeing %d bytes in heap at file %s line %d, heap use now %d bytes\n",
 											 s->size, file, line, state.current_size);
 		checkEyecatchers(file, line, p, s->size);
-		//free(s->ptr);
+		/* free(s->ptr); */
 		free(s->file);
 		state.current_size -= s->size;
 		TreeRemoveNodeIndex(&heap, e, 0);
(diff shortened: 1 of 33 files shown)


icraggs:
When I started writing this client library, it appeared to me that Linux would be the primary OS for embedded devices (this was in the mid-2000s).   This is why it is focused on Linux and WIndows.  The OS dependent functions are generally within specific modules though. The embedded client was started later to be smaller and more portable, but it has less function.

I'd need the Eclipse IP validation for this PR to be fixed before I could merge it.  The error messages are: 

"The following users do not have valid ECAs. Please see this page for details, or to sign a ECA.
rainer.poisel@logicals.com

rainer.poisel@logicals.com"

Thanks

rpoisel:
Thanks Ian for your quick reply. I just signed-off and it looks correctly now.

Would you be interested in a "real" OS layer? I don't know when I could do the necessary change(s), but I think it would increase readability and simplify supporting additional operating systems. Operating system specific functionality would then be resolved during link time, making it possible to avoid almost all `#if` pre-processor directives.

icraggs:
Yes, I would be interested in a "real" OS layer, but I wonder how much advantage it would give us now, given that the current code base works on Linux, Windows and MacOS and is pretty mature?  And there is the embedded client which was intended to be portable and modular.

Perhaps we could start by you outlining your proposed design for an OS layer in an issue?  Then we could discuss further.

Thanks.

cochicde:
Hi. I found this pull request since I also did the adaptation to vxworks 69 and was looking where to push to. I also would like to have a platform layer, in order to make the code easily more expandable. I'd say to do it in a new branch to keep the master safe. I did some platform adaptation for http://git.eclipse.org/c/4diac/org.eclipse.4diac.forte.git/tree/src/arch and the main thing is to define the interface to the platform layer, and then throw the code to each implementation.

icraggs:
I've merged the changes into the develop branch and opened an issue to track the OS layer discussion.  Thanks!

PR #310, opened 2017-07-26
MQTTAsync_global_init exported properly

CallumAttryde:
The function for telling paho to/not to initialise OpenSSL is included in the header (MQTTAsync.h) but not exported from the shared library. This means it can be compiled fine but will run into an undefined symbol error (or similar) at runtime

Signed-off-by: Callum Attryde


icraggs:

PR #317, opened 2017-08-07
Added yield method with sleep patameter

kejsi10:
Publish is taking too long. 
We added a parameterized yield method.


icraggs:
The IP validation on this pull request has failed so that would need to be fixed.  However I'd rather do this so that it was configurable, probably in the global_init call.  Also, we don't need to duplicate the whole function - a shell allowing the sleep time to be altered should be sufficient.

icraggs:
Will add a change for this when fixing #452

PR #341, opened 2017-10-17
Fixup win32 global init

raffienficiaud:
This PR fixes 2 problems:

* the export of the symbols at dll boundary are erroneous: they should be ``dllexport`` when we build the library and ``dllimport`` otherwise. 
  This is now controlled by the ``MQTT_EXPORT`` define (defined when mlibrary being built, not defined otherwise).
  For the object library: since it is built all the time, we can just define the ``MQTT_EXPORT`` to 1.

  See https://cmake.org/cmake/help/v3.0/prop_tgt/DEFINE_SYMBOL.html for more information on how this is done with CMake. 

* the global mutexes creation are now inside a function called ``MQTTAsync_init`` / ``MQTTClient_init``. These functions are called by the ``DLLMain``
  to have the same behaviour as before on Windows and on other platforms, and called also by the currently exposed function ``MQTTAsync_global_init``
  ``MQTTAsync_global_init`` / ``MQTTClient_global_init``

I do not know if the call to ``MQTTAsync_init`` can be done several times without side effects on platforms other than win32. Let me know if you experience problems
with that. 

Thanks


raffienficiaud:
I believe I just signed the eclipse validation. Let me know what I should do next.

raffienficiaud:
Any chance to have a feedback?

marton78:
@icraggs could you please restart the checks? @raffienficiaud says he now signed Eclipse validation.

raffienficiaud:
How should I proceed? Should I resend a PR?

icraggs:
The checks happen automatically, so the best way is to resubmit the PR, yes.

jumoog:
any update on this?

icraggs:
I can't merge these commits without them being signed off: 

raffi.enficiaud@tuebingen.mpg.de

Also, I never use the library statically - I assume both of these changes are for static use?  It would be good to have some regression tests added to the test suite for static use.  That effort would probably be more than these code changes.

raffienficiaud:
@icraggs I updated the PR with signed commits. Let me know if this is what you were expecting.

Concerning the PR itself, 

* The changes are for dynamic and static use.
* `LIBMQTT_API` explicitly indicates the decoration of the functions during the build of the library or when the library is being pulled by another one. The `DllExport`/`DllImport` instead does not make this distinction
* The global mutexes creation is now shared between the global init functions (eg. `MQTTClient_global_init`) and the reasons for attaching the dll to another process - if I remember well.

Let me know if this is understandable. It has been quite some time since...

raffienficiaud:
Should I do anything else?

emgre:
Another thing to consider would be to not compile `DllMain` when doing a static build, as it can create a duplicate symbol error when linking if another library does this error.

cc @jadamcrain

icraggs:

raffienficiaud:
Thanks!

raffienficiaud:
@icraggs should I close the PR or wait?

icraggs:
I was waiting until the Appveyor and Travis builds had completed successfully.   Which they now have.  Thanks again!

icraggs:
@emgre maybe open another issue or PR for that?

keysight-daryl:
this PR is a real improvement in allowing the library to be built as a static library under windows as well as fixing the dll spec.  nice work!

Issue #348, opened 2017-10-30
Macro redefinition (socket.h) in windows10x64


dontsovcmc:
I found the bug. 
No problem with functions and open non blocking connection.

The problem in incorrect defines in socket.h:
```
2>C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\errno.h(43): warning C4005: 'EINTR' : macro redefinition
2>          d:\code\paho.mqtt.c_fork\paho.mqtt.c\src\Socket.h(31) : see previous definition of 'EINTR'
2>C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\errno.h(50): warning C4005: 'EAGAIN' : macro redefinition
2>          d:\code\paho.mqtt.c_fork\paho.mqtt.c\src\Socket.h(29) : see previous definition of 'EAGAIN'
2>C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\errno.h(100): warning C4005: 'ECONNRESET' : macro redefinition
2>          d:\code\paho.mqtt.c_fork\paho.mqtt.c\src\Socket.h(39) : see previous definition of 'ECONNRESET'
2>C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\errno.h(104): warning C4005: 'EINPROGRESS' : macro redefinition
2>          d:\code\paho.mqtt.c_fork\paho.mqtt.c\src\Socket.h(33) : see previous definition of 'EINPROGRESS'
2>C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\errno.h(118): warning C4005: 'ENOTCONN' : macro redefinition
2>          d:\code\paho.mqtt.c_fork\paho.mqtt.c\src\Socket.h(37) : see previous definition of 'ENOTCONN'
2>C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\errno.h(130): warning C4005: 'ETIMEDOUT' : macro redefinition
2>          d:\code\paho.mqtt.c_fork\paho.mqtt.c\src\Socket.h(41) : see previous definition of 'ETIMEDOUT'
2>C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\errno.h(132): warning C4005: 'EWOULDBLOCK' : macro redefinition
2>          d:\code\paho.mqtt.c_fork\paho.mqtt.c\src\Socket.h(35) : see previous definition of 'EWOULDBLOCK'
```
[...]

PR #351, opened 2017-10-30
prevent redefine of variables by VisualStudio 2013

dontsovcmc:
https://github.com/eclipse/paho.mqtt.c/issues/348

### DIFF
--- a/src/Socket.h
+++ b/src/Socket.h
@@ -25,6 +25,9 @@
 #include <ws2tcpip.h>
 #define MAXHOSTNAMELEN 256
 #if !defined(SSLSOCKET_H)
+#if (_MSC_VER < 1900) /* otherwise VS 2013 includes errno.h after our redefine*/
+#include <errno.h>
+#endif
 #undef EAGAIN
 #define EAGAIN WSAEWOULDBLOCK
 #undef EINTR


icraggs:
This pull request seems to contain the changes for SSL options checking, as well as any for variable redefinition.

dontsovcmc:
I commit to develop branch as written in https://github.com/eclipse/paho.mqtt.c/blob/master/CONTRIBUTING.md  :/
Should I commit to master and add defines for VS less than 2017? 

#ifdef _MSC_VER
#if (_MSC_VER < 1910)
#include <errno.h>
#endif
#endif

icraggs:
You shouldn't need to include errno.h - I thought your issue was to avoid the redefinition of constants because of the inclusion of errno.h.  None of this should be needed as I've built versions of this code on Windows for the last 10 years, and I don't think that changing versions of Visual Studio should have such a large effect. 

I installed VS 2013 to try this out, but it seems that it's not compatible with the latest update to Windows 10, so I verify that these changes are not needed.

JuergenKosel:
Hello @icraggs ,

please take a look at https://github.com/eclipse/paho.mqtt.c/pull/367
For Visual Studio 2010 build the current include of errno.h causes a problem in Socket_setnonblocking().

icraggs:
Ok, errno.h should not be included in Socket.c.  It seems I missed that addition in July in a PR #a16cff63 as it was included along with many other changes, and the Appveyor build & test worked.  But the AppVeyor build uses VS2014, and I don't currently use any edition of Visual Studio older than 2015.

I'm going to make myself a coffee :-(

icraggs:
#367 should fix this problem.

dontsovcmc:
yes

Issue #334, opened 2017-09-15
Segfault on missing SSL options


MartinGebert:
Using paho-mqtt-cpp 1.0.0 together with Paho 1.2.0, statically linked. When no `mqtt::ssl_options` are set in the `mqtt::connect_options` object, trying to connect via SSL leads to a segfault in `SSLSocket_createContext()` at line `SSLSocket.c:523`

    if (opts->keyStore)

due to `opts` being `NULL`. Code fragment to reproduce the issue:

    mqtt::client client("ssl://<server>:8883", "<client-id>");
    //mqtt::ssl_options sslOptions;
    mqtt::connect_options options;

    //options.set_ssl(sslOptions);
    options.set_keep_alive_interval(10);
    options.set_clean_session(true);
    ...

    //Crashes in a subthread
    m_client.connect(options);

Expected behaviour: The pointer to the SSL options should be checked for not being NULL, and an error/exception be thrown otherwise.

fpagliughi:
Obviously, the C library shouldn't segfault, but I will also look to see if there was something I could do in the C++ library to help prevent this.

fpagliughi:
I verified this. If you use an _MQTTAsync_connectOptions_ with the default initializer, and fail to set the _ssl_ field, it will segfault if you use a valid "ssl://..." URL. There seem to be a number of places through the library where the ssl field is used, so the check for NULL should probably happen earlier in the connect call.

icraggs:
So the connect call should return an error if the client was created with an ssl:// URI and the sslOptions structure is not created?

fpagliughi:
Yes. If the URI starts with "ssl://" but the `ssl` field/pointer in the connect struct is NULL, then the app crashes with a segfault. It would be better for the connect call to return some type of "bad parameter" error.

dontsovcmc:
I correct this. Please, check SSLSocket_setSocketForSSL return code. I think it should be only 1 if success. https://github.com/eclipse/paho.mqtt.c/pull/350/commits/76f34adceab8d15a051ad6202abcdeab5aa09d0f

icraggs:
I added changes for this issue, along with some tests.  I'm not sure how the PR is supposed to work, that confused me somewhat.

PR #352, opened 2017-10-30
Segfault on missing SSL options

dontsovcmc:
https://github.com/eclipse/paho.mqtt.c/issues/334

Please, check SSLSocket_setSocketForSSL return code. I think it should be only 1 if success. 
It's better to add SSL return code define


icraggs:
I added a fix for #334 which I think is the right way to fix this issue.

PR #357, opened 2017-11-06
Fix stack overwrite

JuergenKosel:
This pull request fixes a stack overwrite, which is caused by bad return value handling
of Socket_writev() in src/Socket.c

### DIFF
--- a/src/Socket.c
+++ b/src/Socket.c
@@ -41,7 +41,7 @@
 #include <string.h>
 #include <signal.h>
 #include <ctype.h>
-
+#include <assert.h>
 #include "Heap.h"
 
 int Socket_setnonblocking(int sock);
@@ -425,6 +425,7 @@ int Socket_writev(int socket, iobuf* iovecs, int count, unsigned long* bytes)
 	}
 #else
 	*bytes = 0L;
+        assert(count > 0);
 	rc = writev(socket, iovecs, count);
 	if (rc == SOCKET_ERROR)
 	{
@@ -433,7 +434,10 @@ int Socket_writev(int socket, iobuf* iovecs, int count, unsigned long* bytes)
 			rc = TCPSOCKET_INTERRUPTED;
 	}
 	else
+        {
+		assert(rc >= 0);
 		*bytes = rc;
+        }
 #endif
 	FUNC_EXIT_RC(rc);
 	return rc;
@@ -473,6 +477,7 @@ int Socket_putdatas(int socket, char* buf0, size_t buf0len, int count, char** bu
 	iovecs[0].iov_base = buf0;
 	iovecs[0].iov_len = (ULONG)buf0len;
 	frees1[0] = 1;
+	assert(count < 5);
 	for (i = 0; i < count; i++)
 	{
 		iovecs[i+1].iov_base = buffers[i];
@@ -482,6 +487,8 @@ int Socket_putdatas(int socket, char* buf0, size_t buf0len, int count, char** bu
 
 	if ((rc = Socket_writev(socket, iovecs, count+1, &bytes)) != SOCKET_ERROR)
 	{
+		if (rc < 0) Log(LOG_SEVERE, -1,"unexpected return value %d of Socket_writev",rc);
+		assert(rc >= 0);
 		if (bytes == total)
 			rc = TCPSOCKET_COMPLETE;
 		else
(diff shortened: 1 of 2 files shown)


src/Socket.c
```diff
@@ -473,14 +476,16 @@ int Socket_putdatas(int socket, char* buf0, size_t buf0len, int count, char** bu
 	iovecs[0].iov_base = buf0;
 	iovecs[0].iov_len = (ULONG)buf0len;
 	frees1[0] = 1;
+	assert(count <= 5);
```
JuergenKosel, line 479: Here is another possible stack overwrite, if the caller sets count > 5 !

src/Socket.c
```diff
@@ -747,6 +752,7 @@ int Socket_continueWrite(int socket)
 	}
 #endif
 
+	assert(5 >= pw->count);
```
JuergenKosel, line 755: Here is another possible stack overwrite, if the caller sets pw->count > 5 !

icraggs:
Thanks Juergen.  Do you actually encounter this situation in real use - if so what are the circumstances?

JuergenKosel:
Hello @icraggs 

I have seen this stack corruption in real use: The broker was disconnected from the LAN while the paho.mqtt.c client has a number of items to publish.

src/Socket.c
```diff
@@ -425,15 +425,18 @@ int Socket_writev(int socket, iobuf* iovecs, int count, unsigned long* bytes)
...
 #else
 	*bytes = 0L;
+        assert(count > 0);
 	rc = writev(socket, iovecs, count);
-	if (rc == SOCKET_ERROR)
+	if (rc < 0)
 	{
 		int err = Socket_error("writev - putdatas", socket);
 		if (err == EWOULDBLOCK || err == EAGAIN)
 			rc = TCPSOCKET_INTERRUPTED;
```
JuergenKosel, line 434: In this line, the return value is set to -22.

src/Socket.c
```diff
@@ -766,7 +772,8 @@ int Socket_continueWrite(int socket)
 		curbuflen += pw->iovecs[i].iov_len;
 	}
 
-	if ((rc = Socket_writev(socket, iovecs1, curbuf+1, &bytes)) != SOCKET_ERROR)
```
JuergenKosel, line 769: Here the return value -22 was not detected as failure.
So the random value in `bytes` was used. And this has lead to stack corruption in a following call of writev().

src/Socket.c
icraggs: writev returns -1 on error, so this shouldn't change anything, right?

src/Socket.c
icraggs: Again, this change will have no effect, that I can see.  (Apart from the assertion, but I don't think we want assertions in production code).

src/Socket.c
icraggs: I don't think the value in bytes will be random - it is initialised to 0 in Socket_writev, and then not written to unless the return from writev was not -1.

icraggs (commented):
The analysis doesn't look right to me.  I think we should start from the symptoms, and work from there.

src/Socket.c
JuergenKosel: In line 433 the return value is changed to -22 (TCPSOCKET_INTERRUPTED), and this value is not handled correctly by the caller!

src/Socket.c
JuergenKosel: Socket_writev() can return -22, and this was not handled here!

The assertions will be disabled, if compiled with macro NDEBUG defined. But for regression tests, this assertions may be useful. At least it is easier to debug a "crash" caused by an assertion, than some random behaviour caused by stack overwrite!

src/Socket.c
```diff
@@ -425,15 +425,18 @@ int Socket_writev(int socket, iobuf* iovecs, int count, unsigned long* bytes)
 	}
 #else
 	*bytes = 0L;
```
JuergenKosel, line 427: I have over seen this line.

JuergenKosel:
I have over seen, that bytes is set to 0 inside Socket_writev().
 So my modification prevent the segfault / stack corruption by accident. Because it prevent correct storing of the pending write.

src/Socket.c
icraggs: I think it is correct, or at least my intention, that the execution path goes into the subsequent code block if the rc is -22 (interrupted).

PR #359, opened 2017-11-10
Visual studio 2013 support

dontsovcmc:
https://github.com/eclipse/paho.mqtt.c/issues/348
For Visual Studio less than 2015.

### DIFF
--- a/src/Socket.h
+++ b/src/Socket.h
@@ -25,6 +25,9 @@
 #include <ws2tcpip.h>
 #define MAXHOSTNAMELEN 256
 #if !defined(SSLSOCKET_H)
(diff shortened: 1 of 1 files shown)


JuergenKosel:
This patch is not necessary, because you could generate visual studio files with cmake for any version you like.

dontsovcmc:
There is a problem with "errno.h" file in VS2013 
"Socket.h" includes before "errno.h",  constants became wrong :( -> socket doesn't works

icraggs:
#367 should fix this problem

dontsovcmc:
yes. I close this.

PR #360, opened 2017-11-10
Segfault on missing SSL options (master branch) 

dontsovcmc:
Commit to master branch.

Please, check SSLSocket_setSocketForSSL return code. I think it should be only 1 if success.
It's better to add SSL return code define


icraggs:
Is this a duplicate of #352?  That was to the master branch too.  However I always add pull requests into the develop branch first, before merging into the master branch.

dontsovcmc:
Yes, I think you ask me to commit to the master branch )... No?

dontsovcmc:
duplicate

PR #367, opened 2017-11-16
Fix Socket_setnonblocking() for Windows (At least if build with MSVC …

JuergenKosel:
…2010)

Partially revert commit a16cff63cc2c595bb9fd0334495953f5d0be7796 for file
src/Socket.c

- Do not include errno.h for Windows build
- Use errno as a local variable in Socket_setnonblocking() for Windows

Signed-off-by: Juergen Kosel <juergen.kosel@softing.com>

### DIFF
--- a/src/Socket.c
+++ b/src/Socket.c
@@ -36,7 +36,9 @@
 #include "SSLSocket.h"
 #endif
 
+#if !defined(WIN32) && !defined(WIN64)
 #include <errno.h>
+#endif
 #include <stdlib.h>
 #include <string.h>
 #include <signal.h>
@@ -99,6 +101,10 @@ int Socket_setnonblocking(int sock)
  */
 int Socket_error(char* aString, int sock)
 {
+#if defined(WIN32) || defined(WIN64)
+	int errno;
+#endif
+
 	FUNC_ENTRY;
 #if defined(WIN32) || defined(WIN64)
 	errno = WSAGetLastError();
--- a/src/Socket.c
+++ b/src/Socket.c
@@ -36,9 +36,6 @@
 #include "SSLSocket.h"
 #endif
 
-#if !defined(WIN32) && !defined(WIN64)
-#include <errno.h>
-#endif
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


JuergenKosel:
For the mqttas library I had an issue with visual studio 2010 builds:
Each connect attempt failed!
Running git bisect shows, that this problem arise by commit a16cff63cc2c595bb9fd0334495953f5d0be7796
This pull request fixes this issue.

icraggs:
errno.h is included in Socket.h, in a section not applying to Windows, so that should not be needed anyway

icraggs:
In fact, I'm guessing that just removing the inclusion of errno.h from Socket.c could fix the problem

src/Socket.c
```diff
@@ -36,7 +36,9 @@
 #include "SSLSocket.h"
 #endif
 
+#if !defined(WIN32) && !defined(WIN64)
 #include <errno.h>
```
JuergenKosel: Should this be completely removed?

src/Socket.c
```diff
@@ -99,6 +101,10 @@ int Socket_setnonblocking(int sock)
  */
 int Socket_error(char* aString, int sock)
 {
+#if defined(WIN32) || defined(WIN64)
+	int errno;
```
JuergenKosel, line 102: This definition of a local variable `errno` is also needed to solve the problem!

src/Socket.c
icraggs: yes

src/Socket.c
icraggs: but this is needed, yes.  It was removed by the VxWorks commit

icraggs:
Do these changes now work for you on VS 2010?

JuergenKosel:
Yes, with this changes mqtt works again for my MSVC 2010 build.

icraggs:
Great!  Thanks.  I'll merge them in.

dontsovcmc:
It works with MSVC 2013, thanks @JuergenKosel  :)

PR #380, opened 2017-11-29
fix issue that MQTTClient_connect hangs on network outage, for windows

ubyyj:
Signed-off-by: ubyyj <wdyuanyoujun@163.com>

On Windows, when there is no network, MQTTClient_connect could hang for very long time. The reason is an negative timeout value could be passed to Thread_wait_sem in Thread.c. On Windows, WaitForSingleObject will be invoked, which expects an DWORD (which is unsigned long), hence the negative timeout value is coverted to an huge DWORD implicitely, hence the thread "hangs".
The fix is: convert negative timeout to zero.

Linux platform does not have the issue, since negative timeout value will not pass the while (++i < count) test, and the function returns immediately.

### DIFF
--- a/src/Thread.c
+++ b/src/Thread.c
@@ -233,7 +233,7 @@ int Thread_wait_sem(sem_type sem, int timeout)
 
 	FUNC_ENTRY;
 	#if defined(WIN32) || defined(WIN64)
-		rc = WaitForSingleObject(sem, timeout);
+		rc = WaitForSingleObject(sem, timeout < 0 ? 0 : timeout);
   #elif defined(OSX)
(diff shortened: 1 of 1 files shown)


ubyyj:
@icraggs would you please take a look?

icraggs:
Thanks.  I'll take a look tomorrow, when I have access to a Windows machine.

ubyyj:
I can easily repro the issue by:
1, with a Windows vm running in VMware Workstation, with network adapter connected
2, start a mqtt client connect to a broker, which will try to reconnect after connection lost
3, disconnect the vm's network for > ~60 seconds (just click the network icon on the bottom right bar of VMware Workstation), 
4, connect vm's network adapter again
you will see the mqtt client won't get connected, log shows function Thread_wait_sem never returns.

icraggs:
Thanks!

Issue #385, opened 2017-11-30
Segmentation faults during disconnect and failed reconnect attempt


JuergenKosel:
Hello,

while continuing testing for issue #373 a segmentation fault was caused in the following line:
https://github.com/eclipse/paho.mqtt.c/blob/c85e4a2fd6e96ebc35caef99a024d94c9b664ba0/src/MQTTAsync.c#L1099

Output from gdb:
```
Program terminated with signal 11, Segmentation fault.
#0  0x0fed5b98 in MQTTAsync_freeCommand1 (command=0x0) at /home/klj/echo-sources-git-svn-full/paho.mqtt.c/src/MQTTAsync.c:1024
1024		if (command->command.type == SUBSCRIBE)
(gdb) up
#1  0x0fed5d50 in MQTTAsync_freeCommand (command=0x0) at /home/klj/echo-sources-git-svn-full/paho.mqtt.c/src/MQTTAsync.c:1054
1054		MQTTAsync_freeCommand1(command);
(gdb) up
#2  0x0fed5f68 in MQTTAsync_writeComplete (socket=26) at /home/klj/echo-sources-git-svn-full/paho.mqtt.c/src/MQTTAsync.c:1105
1105				MQTTAsync_freeCommand(com);
(gdb) bt
#0  0x0fed5b98 in MQTTAsync_freeCommand1 (command=0x0) at /home/klj/echo-sources-git-svn-full/paho.mqtt.c/src/MQTTAsync.c:1024
#1  0x0fed5d50 in MQTTAsync_freeCommand (command=0x0) at /home/klj/echo-sources-git-svn-full/paho.mqtt.c/src/MQTTAsync.c:1054
#2  0x0fed5f68 in MQTTAsync_writeComplete (socket=26) at /home/klj/echo-sources-git-svn-full/paho.mqtt.c/src/MQTTAsync.c:1105
#3  0x0feed364 in Socket_continueWrites (pwset=0x54084bcc) at /home/klj/echo-sources-git-svn-full/paho.mqtt.c/src/Socket.c:822
#4  0x0feeba00 in Socket_getReadySocket (more_work=0, tp=0x54084c8c) at /home/klj/echo-sources-git-svn-full/paho.mqtt.c/src/Socket.c:264
[...]

Issue #373, opened 2017-11-21
Memory leak after QoS 0 messages fill TCP buffer


JuergenKosel:
Hello,

there is a (potential) memory leak in:
https://github.com/eclipse/paho.mqtt.c/blob/c85e4a2fd6e96ebc35caef99a024d94c9b664ba0/src/MQTTPacket.c#L207

Because the memory allocated for buf is not freed, if `Socket_putdatas()` return `TCPSOCKET_INTERRUPTED`.

In the usual case, the corresponding free is in: 
https://github.com/eclipse/paho.mqtt.c/blob/c85e4a2fd6e96ebc35caef99a024d94c9b664ba0/src/Socket.c#L780

But I am unsure, if this memory is freed, if `Socket_close()` calls `ListRemoveItem(s.write_pending, &socket, intcompare);`

JuergenKosel:
I have modified may embedded application to cyclically debug print the memory used by paho.mqtt.c as it is reported in
```
    heap_info* mqtt_mem = Heap_get_info();
    SYS_OUT("MQTT mem current %ld, max %ld",mqtt_mem->current_size,mqtt_mem->max_size);
```

So I see the following output during normal operation:
```
/Firmware/echo/main/CpMainSrv[1042]: EchoMqttMapper_getSentFrames@673:
/Firmware/echo/main/CpMainSrv[1042]: PublishCnt 205906, FailedCnt 0, Pending 25 maxPending 56
/Firmware/echo/main/CpMainSrv[1042]: EchoMqttMapper_getSentFrames@665:
/Firmware/echo/main/CpMainSrv[1042]: MQTT mem current 1904, max 10768
/Firmware/echo/main/CpMainSrv[1042]: EchoMqttMapper_getSentFrames@673:
/Firmware/echo/main/CpMainSrv[1042]: PublishCnt 205906, FailedCnt 0, Pending 25 maxPending 56
/Firmware/echo/main/CpMainSrv[1042]: EchoMqttMapper_getSentFrames@665:
/Firmware/echo/main/CpMainSrv[1042]: MQTT mem current 1904, max 10768
/Firmware/echo/main/CpMainSrv[1042]: EchoMqttMapper_getSentFrames@673:
/Firmware/echo/main/CpMainSrv[1042]: PublishCnt 205906, FailedCnt 0, Pending 25 maxPending 56
```
And when the brocker gets disconnected from the LAN, the amount of memory allocated by mqtt grows (which is expected):
[...]

PR #386, opened 2017-12-04
Test4issue373

JuergenKosel:
This pull request adds a test for issue #373 
This test needs to be started manually.
You need a broker on some other PC / virtual machine, which is cyclically disconnected and reconnected from the LAN.
Then start the test executable, for example with broker address tcp://SLGB-VM-MQTTSRV.inat.internal:1883:
`test/test_issue373 --proxy_connection tcp://SLGB-VM-MQTTSRV.inat.internal:1883`

I was not able to reproduce the segmentation faults reported in #385 with this test. But the memory leak reported in #373

### DIFF
--- a/src/Socket.c
+++ b/src/Socket.c
@@ -685,7 +685,7 @@ int Socket_new(char* addr, int port, int* sock)
 			if (setsockopt(*sock, SOL_SOCKET, SO_NOSIGPIPE, (void*)&opt, sizeof(opt)) != 0)
 				Log(LOG_ERROR, -1, "Could not set SO_NOSIGPIPE for socket %d", *sock);
 #endif
-#if 1
+#if defined(TESTING)
                         {
                             int optsend = 2 * 1440;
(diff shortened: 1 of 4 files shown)


JuergenKosel:
Hello @icraggs 
I have a patch to the library, to force TCPSOCKET_INTERRUPTED for function Socket_new():
```
+#if 1
+                        {
+                            int optsend = 2 * 1440;
+                            if (setsockopt(*sock, SOL_SOCKET, SO_SNDBUF, (void*)&optsend, sizeof(optsend)) != 0)
+				Log(LOG_ERROR, -1, "Could not set SO_SNDBUF for socket %d", *sock);
+                        }
+#endif
```

The question is, should I add this to this pull request?

icraggs:
Thank you very much.  I'm trying out this test to use it or its ideas in automated regression tests.

JuergenKosel:
There is something wrong with the appveyor tests: Adding a new test source file causes tests to fail. Looks like those tests are already broken in the develop branch!

icraggs:
I fixed the AppVeyor tests now - I had to replace the MQTT broker used with the Paho test broker, as the previous one disappeared.

PR #390, opened 2017-12-11
Remove shadow warn. (MQTTAsync_completeConnection)

ghost:
Removes shadow of previous defined local (m).
Especially because the two are of different types:

MQTTAsyncs* m
Messages* m


ghost:
@JuergenKosel mentioned that there might be "... something wrong with the appveyor tests". Could that be the same reason my commit fails appveyor.

icraggs:
Thanks.  (I've fixed the AppVeyor tests now, I had to change the MQTT broker used to the Paho test broker because the old one disappeared)

ghost:
Thank you for all you work! And of course for fixing AppVeyor.

Issue #399, opened 2018-01-03
Separate callback contexts


fpagliughi:
It would be slightly more convenient if each callback function could have its own context. This would allow some of the other languages using the C library to map separate lambdas/closures to each callback. 

It could be done by adding separate callback registration functions, like:
```
MQTTAsync_setConnectionLostCallback(MQTTAsync handle, void* context, MQTTAsync_connectionLost* cl);

MQTTAsync_setMessageArrivedCallback(MQTTAsync handle, void* context, MQTTAsync_messageArrived* ma);

MQTTAsync_setDeliveryCompleteCallback(MQTTAsync handle, void* context, MQTTAsync_deliveryComplete* dc);
```

The existing API wouldn't need to be broken, since the `MQTTAsync_setCallbacks` could just call each of these with the same context value.

icraggs:
Yep.

icraggs:
Fix merged.  Thanks!

PR #403, opened 2018-01-17
Added separate context pointers for each of the global callbacks [#399]

fpagliughi:
This provides a separate context pointer for each of the global callback functions, and provides a means to set each callback individually. This is extremely useful when the library is called from a language that supports lambdas/closures so that each can be individually assigned.


icraggs (commented):
Hi Frank.  To maintain existing behaviour, shouldn't we use the main context pointer (context) in each callback, unless the specific one (e.g. clContext) has been set?

PR #404, opened 2018-01-17
Further lockdown of const strings and buffers in the public Async API

fpagliughi:
Good progress has been made in recent versions of the library to mark some strings and buffers as const. But it had not been completed. This attempts to go through and make the rest of them const.

Basically, I tried to go through the Async API and identify the pointers that were not being used to return values back to the application or were expected to be consumed/freed by the application. So anything pointing to a struct or buffer that is not being modified by the library is marked as const.

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -1566,7 +1566,7 @@ static void MQTTAsync_emptyMessageQueue(Clients* client)
 		{
 			qEntry* qe = (qEntry*)(current->content);
 			free(qe->topicName);
-			free(qe->msg->payload);
+			free((void*) qe->msg->payload);
 			free(qe->msg);
 		}
 		ListEmpty(client->messageQueue);
@@ -1697,7 +1697,7 @@ exit:
 void MQTTAsync_freeMessage(MQTTAsync_message** message)
 {
 	FUNC_ENTRY;
-	free((*message)->payload);
+	free((void*) (*message)->payload);
 	free(*message);
 	*message = NULL;
 	FUNC_EXIT;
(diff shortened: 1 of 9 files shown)


icraggs:
Hi Frank.  I'd be most happy of all if the tests didn't need any const-related changes :-)

fpagliughi:
Most of the changes to the tests were from the code manipulating variables after they were assigned into structs (from the pointers that were now made 'const')

For example it did this:
```
MQTTAsync_connectOptions opts = MQTTAsync_connectOptions_initializer;
MQTTAsync_willOptions wopts = MQTTAsync_willOptions_initializer;

opts.will = &wopts;
opts.will->message = "will message";
opts.will->qos = 1;
```
So it changed the variable struct from the pointer after assigning it. 

I changed it so that the variable was updating the fields...

```
opts.will = &wopts;
wopts.message = "will message";
wopts.qos = 1;
```
To me, this is actually the same thing in terms of setup for the test (the same variable has the same value at the same memory location), just assigned through the variable itself.

I understand that this is effectively a breaking change, but was thinking that the march to MQTT v5 might be a good time to break a few things!

fpagliughi:
Oh, and I did it in several isolated commits in case you wanted to cherry pick which ones you thought were appropriate.

icraggs:
Hi Frank.  There are build problems for this PR.  I tried a couple of the changes and they needed changes in the tests or samples.  I'll look closer this week, but I thought that I'd previously made all the const changes that were possible without causing applications to have to change, so I don't know how many of these will work.

icraggs (changes requested):
Hi Frank.  I'd like to resolve this one way or another.  I'm pretty sure when we added the 'const' modifiers previously I made all the changes possible without affecting existing programs.  Even with the inclusion of MQTT V5, it would require a move to 2.x.x version nos to incorporate changes that could break existing programs, and I didn't feel the need to do that.

It looks like there are some changes that can be incorporated without affecting existing applications, I'll have to sift thru them.  What I do want to avoid is *having* to add 'const' where it wasn't required before.   I hope to get to this next week.

icraggs:
I'm going to close this one now.  There was one last const change you included which didn't affect applications, and that's been added now in #1011.  Thanks.

PR #413, opened 2018-02-08
Fix tokens not set on publish, subscribe, unsubscribe commands 

romansavrulin:
Fixing MQTTAsync client reversed token assignment

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -2585,7 +2585,9 @@ int MQTTAsync_subscribeMany(MQTTAsync handle, int count, char* const* topic, int
 		sub->command.onSuccess = response->onSuccess;
 		sub->command.onFailure = response->onFailure;
 		sub->command.context = response->context;
-		sub->command.token = response->token;
+		/* Replace token to provided in response, if it is valid. */
+		if (response->token)
+			sub->command.token = response->token;
 	}
 	sub->command.type = SUBSCRIBE;
 	sub->command.details.sub.count = count;
@@ -2659,7 +2661,9 @@ int MQTTAsync_unsubscribeMany(MQTTAsync handle, int count, char* const* topic, M
 		unsub->command.onSuccess = response->onSuccess;
 		unsub->command.onFailure = response->onFailure;
 		unsub->command.context = response->context;
-		unsub->command.token = response->token;
+		/* Replace token to provided in response, if it is valid. */
+		if (response->token)
+			unsub->command.token = response->token;
 	}
 	unsub->command.details.unsub.count = count;
 	unsub->command.details.unsub.topics = malloc(sizeof(char*) * count);
@@ -2737,7 +2741,9 @@ int MQTTAsync_send(MQTTAsync handle, const char* destinationName, int payloadlen
 		pub->command.onSuccess = response->onSuccess;
 		pub->command.onFailure = response->onFailure;
 		pub->command.context = response->context;
-		pub->command.token = response->token;
+		/* Replace token to provided in response, if it is valid. */
+		if (response->token)
+			pub->command.token = response->token;
 	}
 	pub->command.details.pub.destinationName = MQTTStrdup(destinationName);
 	pub->command.details.pub.payloadlen = payloadlen;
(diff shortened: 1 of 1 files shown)


icraggs:
Hi.  I think the current code is correct.  The response tokens are created (from the current message id) and set in both the command structure (pub, sub, unsub), and then returned in the token for the response.  You'll see that this change causes tests to fail, because the msgid sent is 0, which is invalid.

romansavrulin:
I've checked the original  code and it completely ignores token set in options, while proposed fix enables this functionality. Maybe we should look into tests or another parts of the code?

romansavrulin:
@icraggs, In [this test log](https://ci.appveyor.com/project/eclipsewebmaster/paho-mqtt-c/build/1.2.218/job/029q2d1r32wkal1k) most errors are related to MQTTClient, which was intact in my commit, not MQTTAsync. And  the last test has failed on connect, not publish.

Travis tests fails indeed, and I've pushed a fix for them.

icraggs:
The token is returned and is the MQTT message id generated by the client library.  It's not intended that the application should choose the token itself.  This would be a change in behaviour, which would open up the possibility of incorrect message id values being chosen.  I don't think it's necessary for applications to need to do this.  For QoS 0, no message id is used in MQTT, so perhaps that is the case you are concerned about?

romansavrulin:
@icraggs according to [hiveMQ documentation](https://www.hivemq.com/blog/mqtt-essentials-part-6-mqtt-quality-of-service-levels) we can use arbitrary message ID generator and even repeat IDs once exchange is over.

image

My app depends on custom ID generator and  Paho.mqtt.c has the ability to utilize this feature, but the assignment of arbitrary ID is done vice-versa in original code, 

```cpp
sub->command.onSuccess = response->onSuccess;
response->token = sub->command.token;
```
instead of

```cpp
sub->command.onSuccess = response->onSuccess;
sub->command.token= response->command.token;
```
and that's why library generated ID is taken over my arbitrary ID, provided via response options. For me, as a user of library, this behavior is not expected.

icraggs:
It's one thing to say what is possible in the MQTT specification, another to say what implementation choices have been made in this library.  

This would be a change in behaviour, even with the 0 check.  I'll have to look at the consequences further.  

I don't think it gives you any advantage to choose your own message id, does it?  How can your app depend on a particular value of the MQTT message id?  That value is lost once it goes through the server.

romansavrulin:
We are implementing mqtt-based logic for device on esp8266 hardware platform, that uses [mqtt-async-library](https://github.com/marvinroger/async-mqtt-client). Because we also use UI components and a lot of other features, the code became too big. So the code-flash_target-test cycle for that platform is about 5-10 minutes per single run and recovery of some exceptions from stack backtrace is painful too (platform lacks hardware debugger).

image


To save time, we've decided to implement all interfaces we use (including some i2c hardware emulations), on windows emulator and succedded a lot in that. Actually, we have the same environment as on target and code with no modifications can be built for both emulator and target platform. In order not to pull lwip stack on emulator, we've decided to use paho.mqtt.c as mqtt backend for [mqtt-async-library](https://github.com/marvinroger/async-mqtt-client). But we cannot use full-featured ASYNC client from paho. Things like multiple subscribe, etc.. is not possible on our target library. It can only do things step-by-step. It operates a single TCP packet to send a message, and then waits for response and calls callback with message id as one of argument. Async nature of the library is used only for power-saving techniques on target device.

Also, target is limited on memory. We have only 40kb for stack and heap (with c++ runtime). So, our logic of working with mqtt is highly depends on those message IDs to get sequential actions. Say, we need to subscribe to 3 topics and after all publish a message. We have to do it step-by-step, and to reference the current step we use message ID. With proposed fix we can set arbitrary message ID to subscribe method and wait for response from server with the same message id to proceed to the next step and so on

romansavrulin:
Oh, forgot to say one important thing. There's a single event callback on the target for the events of the same type, so the only way to see what call triggered particular event is to rely on ID.

icraggs:
Thanks for the explanation.  So as I understand it you are using the Paho client to talk to another client (matt-async-library), rather than with a broker as intermediary?  

You say this is to avoid pulling in the lwip stack on the emulator, but for the Paho client to work you must have some level of TCP support.  Is that more difficult than using a lightweight broker at the back end?

Are you aware of the Paho embedded MQTT client https://github.com/eclipse/paho.mqtt.embedded-c ?  It provides lots of options for just using the pieces you need (just the MQTTPacket library for instance) and putting them together in different ways.

It's still likely that I misunderstood something :-)

PR #428, opened 2018-03-18
Adding TLS 1.3 Compatibility

jaykishanmutkawoa:
Done during IETF101 hackathon together with Nigel Yong (Nigelyong@hackers.mu)

PR #430, opened 2018-03-20
Adding TLS1.3 Compatibility

### LINKED ISSUE #1058: Handling TLS 1.3 connection
Hello,
I would like to connect to a broker which requires TLS 1.3 support. From what I see in the source code, the library supports only TLS 1.0, 1.1 and 1.2. Is it going to change in the near future?

jaykishanmutkawoa:
Signed-off-by: Jaykishan Mutkawoa <jmutkawoa@hackers.mu>

### DIFF
--- a/src/MQTTAsync.h
+++ b/src/MQTTAsync.h
@@ -644,6 +644,7 @@ typedef struct
 #define MQTT_SSL_VERSION_TLS_1_0 1
 #define MQTT_SSL_VERSION_TLS_1_1 2
 #define MQTT_SSL_VERSION_TLS_1_2 3
+#define MQTT_SSL_VERSION_TLS_1_3 4
 
 /**
 * MQTTAsync_sslProperties defines the settings to establish an SSL/TLS connection using the
--- a/src/MQTTClient.h
+++ b/src/MQTTClient.h
@@ -490,6 +490,7 @@ typedef struct
 #define MQTT_SSL_VERSION_TLS_1_0 1
 #define MQTT_SSL_VERSION_TLS_1_1 2
 #define MQTT_SSL_VERSION_TLS_1_2 3
+#define MQTT_SSL_VERSION_TLS_1_3 4
 
 /**
 * MQTTClient_sslProperties defines the settings to establish an SSL/TLS connection using the
--- a/src/SSLSocket.c
+++ b/src/SSLSocket.c
@@ -264,6 +264,9 @@ char* SSLSocket_get_version_string(int version)
 #if defined(TLS3_VERSION)
 		{ TLS3_VERSION, "TLS 1.2" },
 #endif
+#if defined(TLS4_VERSION)
+		{ TLS4_VERSION, "TLS 1.3" },
+#endif		
 	};
 
 	for (i = 0; i < ARRAY_SIZE(version_string_table); ++i)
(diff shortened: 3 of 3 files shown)


jaykishanmutkawoa:
Patch created during the IETF 101 Hackathon togather with Nigel Yong - nigel@hackers.mu

jaykishanmutkawoa:
Hello @icraggs , I already signed the agreement as asked in https://github.com/eclipse/paho.mqtt.c/pull/428 
Can you have a look please ?

icraggs:
There's no test, so I'd like to know out of interest, did you try out this change and have it working?

jaykishanmutkawoa:
@icraggs  We tested it. See this link https://github.com/eclipse/paho.mqtt.c/pull/428/commits/0bbddd1d499274c7ff7213863280b9a84540be30

icraggs:
Ok, I was just wondering how I could test it, so that a test could be added to the automated suite.  Is there some particular Mosquitto configuration for instance?  The test broker I am using is written in Python, so Python configuration would also be good.

icraggs (changes requested):
I will get to looking at TLS version support.  I can't merge this particular PR unless the ECA is signed.  Tests will need to be udpated too.

mtrensch:
IMHO this PR will default to accept TLS 1.3 brokers only, as the min and max protocol versions are both fixed to TLS_VERSION1_3, independent of the settings in sslVersion and for all sockets.

mtrensch:
For running your internal tests using https://github.com/eclipse/paho.mqtt.testing.git you will need to add TLS 1.3 to the broker by changing SSLContext in "interoperability/mqtt/brokers/listeners/TCPListeners.py" as follows
```
diff --git a/interoperability/mqtt/brokers/listeners/TCPListeners.py b/interoperability/mqtt/brokers/listeners/TCPListeners.py
index 4d9f237..7bc2f2a 100644
--- a/interoperability/mqtt/brokers/listeners/TCPListeners.py
+++ b/interoperability/mqtt/brokers/listeners/TCPListeners.py
@@ -249,7 +249,7 @@ def create(port, host="", TLS=False, serve_forever=False,
     bind_address = host
   server = ThreadingTCPServer((bind_address, port), WebSocketTCPHandler, False)
   if TLS:
-    context = ssl.SSLContext(protocol=ssl.PROTOCOL_TLSv1_2)
+    context = ssl.SSLContext(protocol=ssl.PROTOCOL_TLS)
     try:
       context.set_ciphers('ALL:@SECLEVEL=1') # until we have seclevel 2 TLS config
     except:
```

**NOTE:** Debian buster (what I am using) seems to have disabled TLS 1.0 and 1.1, thus the broker will only run with TLS 1.2+ clients. See https://bugs.python.org/issue31453

I messed around in the code a little to also provide a test case, but I don't like it, as I needed to introduce an API function to get the used TLS Version (I don't know of any other way how to acquire this information). See https://github.com/mtrensch/paho.mqtt.c/commit/52b5aee82e9b80e20f96ddaa194768388572ea4f

**BTW** I think when using openssl 1.1+ the defaults are already used and TLS 1.3 is established, as TLS_version_method() is then used: https://github.com/eclipse/paho.mqtt.c/blob/64a5ff3c3b71fe019353aeacaebc66a3cf4f3461/src/SSLSocket.c#L554 But at current time you cannot force TLS version being used, but accept all supported TLS versions by openssl

dns2utf8-novaziun:
Hi all
Is this PR still going forward?
Regards

CDMartinL:
I think this was superseded by #1641.

PR #438, opened 2018-04-05
Implement an strerror-like helper function

dawagner:
Returns a human-redable error string.

Signed-off-by: David Wagner <david.wagner@easymile.com>


dawagner:
I'll be re-creating this PR against master instead.

PR #451, opened 2018-05-11
Fix typo in printf format string

Thalhammer:
Fix a type in printf format string which could lead to out of bounds memory access on shutdown.
Signed-off-by: Dominik Thalhammer <dominik@thalhammer.it>


icraggs:
Thanks!

PR #454, opened 2018-05-16
Add separate contexts for each callback

### LINKED ISSUE #399: Separate callback contexts

Thalhammer:
Fixes #399 

Signed-off-by: Dominik Thalhammer <dominik@thalhammer.it>

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -298,9 +298,11 @@ typedef struct MQTTAsync_struct
 
 	/* "Global", to the client, callback definitions */
 	MQTTAsync_connectionLost* cl;
+	void* cl_context; /* the context to be associated with the connectionLost callback*/
 	MQTTAsync_messageArrived* ma;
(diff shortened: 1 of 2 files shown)


icraggs:
Hi.  Thanks for this, but I merged #403 for the same function.

PR #511, opened 2018-08-09
Coverity resource leaks

ralight:


icraggs:
Thanks Roger.  You're on the ball - I only submitted the coverity build late yesterday! I wasn't expecting you to notice.

ralight:
Coverity sent me an email and I was looking at it for Mosquitto as well, so I thought I'd take a peak. Those were the easy ones to fix!

PR #521, opened 2018-08-16
Add OpenSSL error handler callback function and context

### LINKED ISSUE #326: Suppress SSL error dumping
It would be nice to have a macro which hides the call to ERR_print_errors_fp() in SSLSocket.c. I found that although this was useful for development debugging, in practice/production I did not want stderr to be polluted uncontrollably. An option to disable this at build time (rather than maintaining divergent patches masking functionality) would be much appreciated

tbeu:
* Add new members to SSL options structures MQTTClient_SSLOptions / MQTTAsync_SSLOptions to set an application-specific OpenSSL error handler callback function and context
* Increment the structure version and fall back to nullified error handler, i.e., no more OpenSSL messages are printed to stderr
* This removes the error reporting for SSLSocket_getch, SSLSocket_getdata, SSLSocket_putdatas and SSLSocket_continueWrite where the SSL options structure is not trivially available

Close #326

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -2702,7 +2702,7 @@ int MQTTAsync_connect(MQTTAsync handle, const MQTTAsync_connectOptions* options)
 	}
 	if (options->struct_version != 0 && options->ssl) /* check validity of SSL options structure */
 	{
-		if (strncmp(options->ssl->struct_id, "MQTS", 4) != 0 || options->ssl->struct_version < 0 || options->ssl->struct_version > 2)
+		if (strncmp(options->ssl->struct_id, "MQTS", 4) != 0 || options->ssl->struct_version < 0 || options->ssl->struct_version > 3)
 		{
 			rc = MQTTASYNC_BAD_STRUCTURE;
 			goto exit;
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -2859,6 +2859,11 @@ int MQTTAsync_connect(MQTTAsync handle, const MQTTAsync_connectOptions* options)
 			if (options->ssl->CApath)
 				m->c->sslopts->CApath = MQTTStrdup(options->ssl->CApath);
 		}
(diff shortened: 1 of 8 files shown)


icraggs:
You need to change the check of the SSL options structure version in MQTTAsync_connect, line 2705 in MQTTAsync.c, otherwise the TLS tests won't work.   Oh - and the equivalent in MQTTClient.c.

icraggs:
And any chance we could get small diffs on the header files?   All the spacing has changed, which is not helpful for anyone reviewing the changes.  Thanks!

tbeu:
> And any chance we could get small diffs on the header files?

Sorry. I reverted and force pushed by f2b8c314bdf1f0995b681e678c59536c967642ac.

> You need to change the check of the SSL options structure version in MQTTAsync_connect, line 2705 in MQTTAsync.c, otherwise the TLS tests won't work. Oh - and the equivalent in MQTTClient.c.

Right, I missed that one. Fixed by 98cd807b6302e8df47f24db0ce7485968cfd5d05.

icraggs:
Thanks.  One last thing.  I'd like to see a test or sample program before merging, to show that it works.  We could update one or more of the samples, like paho_c_pub.

tbeu:
This PR solves the OPENSSL_Applink issue with ERR_print_errors_fp on Win, if you build a DLL wrapper around paho-mqtt3cs.dll/paho-mqtt3as.dll and also use the OpenSSL libraries as DLLs. See https://stackoverflow.com/q/24130476/8520615 or http://openssl.6102.n7.nabble.com/OPENSSL-Uplink-no-OPENSSL-Applink-td46124.html for some details.

tbeu:
We still have the option to only remove the call to ERR_print_errors_fp and to not bloat the structure. Whatever you prefer.

icraggs:
I'm happy to go with the change.  I've merged it into develop locally and was just doing some testing with it.  I am planning to include it with the release.

PR #533, opened 2018-09-05
Fix Makefile install target

RasmusAntons:
I am not sure if the html target should be called here, but somehow the manpages for MQTTClient.h.3 and MQTTAsync.h.3 need to be generated (or excluded from the install).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,7 +269,7 @@ strip_options:
 
 install-strip: build strip_options install
 
-install: build
+install: build html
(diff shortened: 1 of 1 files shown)


jplflyer:
The other choice is to update the README.md file to indicate one should make the html before make install.

icraggs:
A fix has been added which means the install works if the man pages have not been generated, so you can choose.  Thanks.

PR #537, opened 2018-09-10
Fix build on Cygwin

gitplcc:
Fix no case for endian.h inclusion in src/SHA1.c under Cygwin

Signed-off-by: Pedro Luis Castedo Cepeda <pedroluis.castedo@upm.es>


icraggs:
Do you use CMake to build on cygwin?

PR #539, opened 2018-09-12
Prevent segfault by current uri

JuergenKosel:
This merge request solves a segmentation fault issue, that occur during connect attempts.


icraggs:
Thanks.

Issue #545, opened 2018-09-19
Flexible yet secure SSL/TLS version


janschatz:
Paho doesn't seem to allow me to enable several SSL/TLS versions unless I chose MQTT_SSL_VERSION_DEFAULT. I don't want to use MQTT_SSL_VERSION_DEFAULT, because it enables SSLv2 and SSLv3, which are both not considered secure anymore. However I also don't want to limit my users to only one of TLSv1.0, TLSv1.1 and TLSv1.2. I want to have them all, because all of them are considered secure at the moment.
Do I have any options?

tbeu:
One option for the library would be to disable SSLv2 permanently and only enable SSLv3 conditionally at compile time. This is the way how OpenSSL considers weak security.

icraggs:
We'll have to decide what the right TLS versions are appropriate now.  1.3 should probably be part of the mix too.   I expect testing to be the most time consuming aspect.

tbeu:
#430 deals with TLS v1.3.

icraggs:
Yes, #430 deals with the changes needed for TLS v1.3, but not the testing :-)

janschatz:
I was more asking from a user's point of view. So I understand there currently is no way to achieve this.

How about adding something like `MQTT_SSL_VERSION_TLS_1_0_OR_HIGHER` etc.? Typically I want to have newer and more secure versions automatically (i.e. with new OpenSSL versions) and just disable older versions on purpose. The creation of the context could look like this in `SSLSocket_createContext`:
```
net->ctx = SSL_CTX_new(TLS_client_method());

switch(sslVersion)
{
  ...
  case MQTT_SSL_VERSION_TLS_1_0_OR_HIGHER:
    SSL_CTX_set_options(net->ctx, SSL_OP_NO_SSLv3);
    // Fall through
  case MQTT_SSL_VERSION_TLS_1_1_OR_HIGHER:
    SSL_CTX_set_options(net->ctx, SSL_OP_NO_TLSv1);
  ...
}
```

a-andre:
Support for TLS 1.0 should be dropped, see https://blog.pcisecuritystandards.org/are-you-ready-for-30-june-2018-sayin-goodbye-to-ssl-early-tls

People depending on that version can use an older version of paho.

PR #550, opened 2018-09-24
Flexible yet secure SSL/TLS version

janschatz:
Add SSL/TLS version options MQTT_SSL_VERSION_TLS_1_0_OR_HIGHER, MQTT_SSL_VERSION_TLS_1_1_OR_HIGHER and MQTT_SSL_VERSION_TLS_1_2_OR_HIGHER, which can be used to disable outdated versions, but will include future versions
See #545

### DIFF
--- a/src/MQTTAsync.h
+++ b/src/MQTTAsync.h
@@ -889,6 +889,9 @@ typedef struct
 #define MQTT_SSL_VERSION_TLS_1_0 1
 #define MQTT_SSL_VERSION_TLS_1_1 2
 #define MQTT_SSL_VERSION_TLS_1_2 3
+#define MQTT_SSL_VERSION_TLS_1_0_OR_HIGHER 4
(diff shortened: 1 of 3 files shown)


janschatz:
Well, seems like I need some help with this signed-off-by stuff. I forgot to add it to the commit message.

loganaden:
we will send an updated patch for TLS 1.3 based on https://github.com/eclipse/paho.mqtt.c/pull/430

janschatz:
In my opinion my pull request is independent of your TLS 1.3 pull request (of course they should be synchronised). It just introduced the option to say "I want AT LEAST TLS 1.1".

icraggs:
Thanks for the contribution and sorry for the delay.  One of the commits didn't include the git 'signed-off-by' info.  If that isn't fixed I'll look to add these updates myself.

PR #517, opened 2018-08-16
Fix MSVC build errors with Visual Studio 2010

tbeu:

PR #560, opened 2018-10-10
Fix compilation with VS2010

AdrianMoranMontes:
Until VS2013, all variables should be defined at the beggining of a block. This patch solves the issue.

### DIFF
--- a/src/Log.c
+++ b/src/Log.c
@@ -3,11 +3,11 @@
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
- * and Eclipse Distribution License v1.0 which accompany this distribution. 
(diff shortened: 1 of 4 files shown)


AdrianMoranMontes:
I see travis-ci is failing on test 40 when compiling with clang. I can not reproduce such failure in my local environment, since the test successes both, with GCC and with CLang.

Can anybody try to reproduce the error?

tbeu:
See also #517.

tbeu:
See also #517.

icraggs:
Thanks.

tbeu:
I am rather confused why you merged this outdated and incomplete PR after I wrote in https://github.com/eclipse/paho.mqtt.c/pull/517#issuecomment-558790997 that #517 supersedes this PR.

PR #561, opened 2018-10-10
Fix compilation for Windows 32 bit systems

AdrianMoranMontes:
Functions htonll and ntonll seems not to be available on Windows 32 bits system. This workaround perform that functionallity swipping bits.

### DIFF
--- a/src/WebSocket.c
+++ b/src/WebSocket.c
@@ -47,10 +47,15 @@
 #  if BYTE_ORDER == LITTLE_ENDIAN
 #    define htobe16(x)   htons(x)
 #    define htobe32(x)   htonl(x)
+#if defined(WIN32)
+#    define htobe64(x)   ((1==htonl(1)) ? (x) : ((uint64_t)htonl((x) & 0xFFFFFFFF) << 32) | htonl((x) >> 32))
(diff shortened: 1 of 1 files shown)


icraggs:
Hi.  We build with 32 bit Windows all the time, on Visual Studio 2013:

https://ci.appveyor.com/project/eclipsewebmaster/paho-mqtt-c/builds/18661028

so I think it's due to the version of Visual Studio you are using.  If these definitions have been copied from elsewhere, we need to know the source, and the license.  If the license is GPL, we can't use them as that license conflicts with the EPL.

AdrianMoranMontes:
Hi @icraggs ,
in first place, thank you for your reply.

I forgot to mention, I am compiling using VS2010 framework.

Regarding to the license, I took the solution from here:

https://github.com/esnet/iperf/blob/master/src/portable_endian.h

Which claims to be "public license". This definitions also seem to be present on NetBSD repositories, which have more restrictive license.

Thank you!

CapitanFuture:
Your patch help me to compile with MinGW 5.3.0 on a Win10 System! 
Hopefully they accept your pull request !

tbeu:
See also #517.

icraggs:
This has been fixed in #517, using Windows library functions so we don't have the IP question.  Thanks.

Issue #28, opened 2016-02-04
Callback onSuccess or onFailure not invoked from MQTTAsync client


jpwsutton:
migrated from Bugzilla [#444103](https://bugs.eclipse.org/bugs/show_bug.cgi?id=444103)
status RESOLVED severity _normal_ in component _MQTT-C_ for _1.1_
Reported in version _1.0_ on platform _PC_
Assigned to: Ian Craggs

Original attachment names and IDs:
- _[file_444103.txt](https://bugs.eclipse.org/bugs//attachment.cgi?id=247065)_ (ID 247065)

On 2014-09-15 08:20:53 -0400, Juergen Denner wrote:

> Created attachment 247065
> Paho trace without callback invocation
> 
> In my scenario I need to do a cleanup on the provided context for callback methods. Therefore it is crucial that one of the callback methods is invoked by the MQTTAsync.c client. 
> 
> A trace is attached that documents the following flow:
> 1. SEND. TOPIC="test",  qos = 1, Payload = "Msg before disconnect. ...", Payload length = 144
> 2. Immediate disconnect with timeout = 0
> 
> Line 152 shows the freeing of the message. No callback was executed.

On 2014-09-16 11:47:44 -0400, Ian Craggs wrote:

> Ah, I see.  Calling disconnect before waiting for the publish to complete.  Presumably you connected cleansession=true?  If we had connected cleansession=false then the message could complete at a later stage, when the client reconnects.

On 2014-09-17 03:09:23 -0400, Juergen Denner wrote:

> Correct, cleansession was set to true.

On 2015-08-10 17:54:55 -0400, Ian Craggs wrote:

> So finally I am thinking about addressing this bug.  There are a number of cases.
> 
> 1) Disconnect clean session while there are outstanding publish requests.  We should call onFailure, it is clear.
> 
> 2) Disconnect non clean session, while there are outstanding publish requests.  These could be completed on reconnect, so we should not make any callback calls at this point (no change needed).
> 
> 3) The client object is destroyed while there are outstanding publish requests.  This implies a previous disconnect non clean session.  If persistence is enabled, the client objects could be recreated at a later date, and the publish requests continued.  However the callbacks would not be re-instantiated then, but the global callbacks messageArrived() and deliveryComplete() would be called appropriately.   I propose calling onFailure, with an "interrupted" or similar return code.

On 2015-08-10 23:50:29 -0400, Frank Pagliughi wrote:

[...]

PR #562, opened 2018-10-10
Remove unneeded variable declaration.

AdrianMoranMontes:
IMHO, it's clearest with the direct casting. Moreover, it avoid warnings/errors with certain compilers.


AdrianMoranMontes:
More information:

Some compilers raise an error if this change is not applied. In my case, I tried to compile with GHS (Green Hills Software) compiler and the following error was present:

`error #28: expression must have a constant value`

Issue #146, opened 2016-06-18
Add pre-shared key (PSK) support


icraggs:

ralight:
You can run `grep -r TLS_PSK lib/*` on the mosquitto source to find how it's done there.

It's basically just defining a callback and configuring it with `SSL_CTX_set_psk_client_callback()`.

icraggs:
Thanks Roger, I was thinking of asking :-)

tenderitaf:
Hello, as a reference for the others 

You may want to integrate paho with a JSSE lib from google called conscrypt that implements PSK.

https://github.com/google/conscrypt

Usage example from their integration test suite ;

https://github.com/google/conscrypt/blob/f0e05d561147458e3a10b4806e24d0752dfd5d4b/openjdk-integ-tests/src/test/java/libcore/javax/net/ssl/SSLSocketTest.java

Regards

icraggs:
Done in #573

PR #573, opened 2018-10-30
Add support for TLS-PSK

janderholm:
This change adds support for TLS-PSK. It is done by exposing the OpenSSL callback almost directly to the client so that they may use the _hint_ configured by the server if wanted. It supports registering a application specific context pointer to the callback.

I have also implemented a simple callback to one of the sample apps. Please tell me if I should add it to the others as well.

The sample application have been tested with mosquitto 1.4.14 configured with `psk_file` and `psk_hint`. Note that it is important to specify `psk_hint` or mosquitto will ignore TLS-PSK.

Solves  #146


janderholm:
I have now signed the ECA. But I don't know how to rerun the checks?

icraggs:
Thanks. If you add another commit, then I think the checks will be re-run.  The CI tests have failed on the TLS tests, so I expect something in your changes is causing those to fail.

janderholm:
I have added two tests for PSK. They work fine locally, but not on Travis. I think this is because Travis does not use mosquitto.conf from this package so I made a pull request here:
https://github.com/eclipse/paho.mqtt.testing/pull/57

janderholm:
I have updated the reviews to include the ability to disable loading of the openssl default certs. The reason is when doing PSK mutual authentication it is probably best to not accept any other brokers even though they may have properly signed certificates.

icraggs:
The tests don't work on Travis because the CI tests there are using the Paho test broker rather than Mosquitto.  I've started looking at adding PSK support to that broker, but in the meantime I'll probably just disable the PSK tests.

Issue #588, opened 2018-12-04
Persistence key for command is the same for v5 and basic


scollon42:
Hi!
Thanks for your awesome job with Paho!

I was investigating an issue I had with persistence today and I realize that 2 const in `MQTT_Persistence.h` have the same value.

```c
/** Stem of the key for an async client command */
#define PERSISTENCE_COMMAND_KEY "c-"
/** Stem of the key for an MQTT V5 async client command */
#define PERSISTENCE_V5_COMMAND_KEY "c-"
```

I don't know if this is the correct behavior but when I change to :
```c
#define PERSISTENCE_V5_COMMAND_KEY "c5-" 
``` 
and it help with my issue.

To develop on the issue I'm talking about, i was trying test paho behavior if some, to be published, messages are persisted and then the local application is killed and restarted. And the messages stored in the filesystem where not restored.

I'm new with MQTT and paho and maybe I missed some information but I wanted to open an issue (I'll probably create a merge request too) in case it needed to be fixed!

Thanks!

icraggs:
Thank you!  Yes, looks like a copy and paste error.

PR #589, opened 2018-12-04
Update PERSISTENCE_V5_COMMAND_KEY value

scollon42:
update value from c- to c5- to differentiate it from PERSISTENCE_COMMAND_KEY
fix for issue #588


icraggs:
Thanks.  I had to fix a number of related issues that were highlighted by this change too.

scollon42:
Thanks for your work!

PR #591, opened 2018-12-05
Adding support for custom HTTP headers when using WebSocket.

gyf304:
Signed-off-by: Yifan Gu <gyf304@gmail.com>

Custom / additional HTTP headers are sometimes needed for MQTT services behind a reverse proxy, usually for authentication, but can also be used to carry additional metadata that sometimes need to be provided to the server.

For example, for the AWS IoT WebSocket MQTT broker, when using custom authorizers, need specific headers set before establishing a WebSocket connection.

See https://docs.aws.amazon.com/iot/latest/developerguide/custom-auth.html

The current implementation of WebSocket does not allow adding custom / additional headers. This patch adds that feature.

### DIFF
--- a/src/Clients.h
+++ b/src/Clients.h
@@ -85,6 +85,7 @@ typedef struct
 #endif
 	int websocket; /**< socket has been upgraded to use web sockets */
 	char *websocket_key;
+	const MQTTClient_nameValue* httpHeaders;
 } networkHandles;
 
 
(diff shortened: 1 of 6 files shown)


icraggs:
Thanks very much.  Looks good.  Ideally we'll want a test to validate too.

gyf304:
I'll try the add a test over the weekend.

icraggs:

Issue #598, opened 2018-12-28
CMake Android build links against pthread


asleszyn:
CMake Android build fails during linking because the linker cannot find -lpthread. In the NDK, pthread is included in libc so there's no need to link against it.

icraggs:
This should be fixed by the PR.

PR #612, opened 2019-02-05
CMake Android build fails during linking because the linker cannot fi…

tatsuya-ogawa:
CMake Android build fails during linking because the linker cannot find -lpthread. In the NDK, pthread is included in libc so there's no need to link against it.
Related issue is #598.


icraggs:
Thanks.  For me to merge this automatically, you'll need to sign the Eclipse ECA (if you haven't already) and sign-off the commits.

DoumanAsh:
Will it get merged any time soon?

On side note I wonder if you could just use `find_package(Threads)` instead of ifdef per OS

PR #613, opened 2019-02-14
Added missing verification for SSL server's IP address

rmbbgh:
The PAHO SSL code was checking only DNS names from certificate's Subject Alternative Name.
Thus if the "verify" was properly set to 1 in the MQTTClient_SSLOptions and connection was made
to the IP address (not the DNS name), the SSL verification was failing regardless the server
had valid cert.
Samples incorrectly assumed default value for verify as 1, so never set it to 1, only reset to 0.
Thus samples were always operating as if --insecure flag was specified.
This is fixed in this commit as well.

Signed-off-by: Maksym Ruchko <mruchko@advantech-bb.com>


icraggs:
Thanks!

Issue #194, opened 2016-11-26
Support CMake configure for apps using paho 


fmikulu:
You can find attached patch and a simple **Findpaho.cmake** finder, resp. **pahoConfig.cmake.in**
The finder can be saved somewhere in a project referencing paho library, e.g:

_referencing-app/cmake/modules/FindPaho.cmake_

And apply commands in its CMakeLists.txt:

**set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
FIND_PACKAGE(paho REQUIRED)
IF (PAHO_FOUND)
    MESSAGE(STATUS "Paho found: " ${PAHO_ROOT_DIR} )
ELSE()
    MESSAGE("Paho cannot be found")
ENDIF()**

Another way is to incorporate **pahoConfig.cmake.in** to paho source root and call there from CMakeLists.txt:

**configure_file(${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_NAME}Config.cmake.in  {PROJECT_BINARY_DIR}/${PROJECT_NAME}Config.cmake @ONLY)**

fmikulu:
And the attachment ;-)
[modules.zip](https://github.com/eclipse/paho.mqtt.c/files/614294/modules.zip)

icraggs:
Fix added from PR #636.

PR #636, opened 2019-04-04
Export CMake targets

ivoshopov:
In order to use Eclipse-Paho-MQTT-C library, we need to include its headers (MQTTAsync.h, ...). But we need to know where these headers are installed. CMake setup that path (for example -I/usr/local/include) automatically through eclipse-paho-mqtt-cConfig.cmake. This patch generates the Config.cmake base on TARGET_INCLUDE_DIRECTORIES.
fmikulu suggest the same things in issue #194 with a little bit different implementation.
rpoisel also suggest replacing the include_directories() in issue #181

eclipse-paho-mqtt-cConfigVersion.cmake is added according to
https://cmake.org/cmake/help/v3.14/manual/cmake-packages.7.html#package-version-file

Any suggestions are welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@
 
 # Note: on OS X you should install XCode and the associated command-line tools
 
-CMAKE_MINIMUM_REQUIRED(VERSION 2.8.4)
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8.12)
 PROJECT("Eclipse Paho C" C)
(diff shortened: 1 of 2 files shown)


icraggs:
I'm no CMake expert so the main thing I'm concerned about is making sure the existing build still works (which it does) and that there is least impact on all the other ways of building with the CMake options (static library, etc).  I'm not sure how to best assuage that last concern, but we can merge and then see what happens.

PR #643, opened 2019-04-17
Add disconnect option initializer for mqttv5

JGrossholtz:
New fields where added to the MQTTAsync_disconnectOptions structure for
MQTTv5 support. The macro provided to initialize this macro, does not
included the new fields. Depending on compilation options it may cause a
warning similar to: 'warning: missing initializer for field
‘onSuccess5’".

Add a MQTTAsync_disconnectOptions_initializer5 with the two missing
fields (onSuccelss5 and onFailure5) set to NULL. Update the previous
initializer and set the struct_version to 0 as it should not be used for
MQTTv5 anyway.

### DIFF
--- a/src/MQTTAsync.h
+++ b/src/MQTTAsync.h
@@ -1227,8 +1227,11 @@ typedef struct
 	MQTTAsync_onFailure5* onFailure5;
 } MQTTAsync_disconnectOptions;
 
-#define MQTTAsync_disconnectOptions_initializer { {'M', 'Q', 'T', 'D'}, 1, 0, NULL, NULL, NULL, MQTTProperties_initializer, MQTTREASONCODE_SUCCESS }
+#define MQTTAsync_disconnectOptions_initializer { {'M', 'Q', 'T', 'D'}, 0, 0, NULL, NULL, NULL,\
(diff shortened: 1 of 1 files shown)


JGrossholtz:
I believe I should add an automated test for this. Where should I do it ?

icraggs:
Hi Julien.  Thanks for the fix, and your offer of an automated test.  I'm not sure it's worth the effort of adding a test, as this type of issue is likely to stay fixed once it is.  

Also, how would you test it?  Add the initialisation check option to the build, and then look at the output?  Or another way?

JGrossholtz:
Hi Ian,

I would use MQTTAsync_disconnectOptions_initializer5 instead of MQTTAsync_disconnectOptions_initializer in some existing tests to ensure we never get a segfault or a different behavior because one initializer set a state the other doesn't. But I don't know how your tests are working. Let me know what you think.

icraggs:
I expect the existing tests are working by accident.  If they had failed, I would have fixed the problem by now!

Correcting the existing tests would be great.  If you can see a need for adding a new test, then I'm all in favour.   I am now going to be on vacation until the 29th April - so you're unlikely to see a further response until then.

PR #654, opened 2019-05-07
Fix symlink location for synchronous shared library

xandrew25:
The symlink must be created in / removed from the library directory.


icraggs:
Thanks.

PR #664, opened 2019-05-30
Patch for mingw compile

CapitanFuture:
Ok next try with a valid ECA id ;)

with this additions it's fine to compile with mingw under a win32 system


CapitanFuture:
I'm lost in space, with that legal eclipse stuff :( I don't want to commit some small stuff here any longer !

PR #675, opened 2019-07-10
Mqtt client cycle crash

michalkas:
The crash happens when during execution of MQTTClient_run method, just before 
MQTTClient_cycle (which is not protected by the mutex) MQTTClient_destroy method is triggered.
It will crash inside MQTTClient_cycle.


icraggs:
Superseded by #676

PR #676, opened 2019-07-11
Fix crash in MQTTClient_cycle when MQTTClient_destroy executed during MQTTClient_run 

michalkas:
The crash happens when during execution of MQTTClient_run method, just before 
MQTTClient_cycle (which is not protected by the mutex) MQTTClient_destroy method is triggered.
It will crash inside MQTTClient_cycle.

Signed-off-by: Michal Kasperek <michalkasperek@gmail.com>

### DIFF
--- a/src/MQTTClient.c
+++ b/src/MQTTClient.c
@@ -2281,84 +2281,90 @@ static MQTTPacket* MQTTClient_cycle(int* sock, unsigned long timeout, int* rc)
 		tp.tv_usec = (timeout % 1000) * 1000; /* this field is microseconds! */
 	}
 
-#if defined(OPENSSL)
-	if ((*sock = SSLSocket_getPendingRead()) == -1)
+	do
 	{
-		/* 0 from getReadySocket indicates no work to do, -1 == error, but can happen normally */
-#endif
-		*sock = Socket_getReadySocket(0, &tp, socket_mutex);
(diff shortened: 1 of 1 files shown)


icraggs:
Is to fix the same issue as #675?  As is stands, it causes multiple tests (mostly concerning connection lost) to fail.

michalkas:
Yes, it is the same issue. Just didn't know how to abandon that pull request. 
I've noticed that this fix is not sufficient, it makes PAHO mqtt to work really slow.

We've came up with another solution. I will propose it next week when I'm back from holidays.

icraggs:
You need to be careful when calling destroy, that the client has actually been disconnected properly.  A trace of the situation (as described in the readme) could help diagnose.

michalkas:
I will add the trace next week if it helps to diagnose it better. 
More detailed issue description is below:
The crash happens even though we are not connected. We are starting the application which tries to connect and we are closing it immediately afterwards. In this case we are not connected yet.
If it happens that during execution of the MQTTClient_run just after unlocking mqttclient_mutex the MQTTClient_destroy method is called the MQTTClient_cycle will work on released resources which causes the app to crash.

icraggs:
It sounds like a simple test program will do the job then - that would be ideal as it can be added to the CI tests.

j8phi:
I think I'm having the same issue. Did you resolve it or working on a new fix?

PR #686, opened 2019-07-23
Heap memset to avoid uninitialized memory warnings in valgrind

dpostorivo:
After running test against the development branch yesterday on a new laptop, I noticed Valgrind was complaining about uninitialized memory when checking incoming mqtt messages. 

The origin of the uninitialized memory was from mymalloc in Heap.c. I just added simple memsets to set all allocated memory to 0 before use.


icraggs:
Thanks

Issue #572, opened 2018-10-26
Segmentation fault when sending a lot of messages from two threads


heing:
Running the following code will after some time produce a segmentation fault inside the while loop in MQTTAsync_countBufferedMessages.

```
#include <MQTTAsync.h>
#include <iostream>
#include <thread>
#include <mutex>

static bool terminated = false;

std::mutex mutex;

static MQTTAsync CreateClient(std::string client_id) {
  MQTTAsync_init_options inits = MQTTAsync_init_options_initializer;
  MQTTAsync_global_init(&inits);

  MQTTAsync mqtt;
  MQTTAsync_createOptions create_options = MQTTAsync_createOptions_initializer;
  create_options.sendWhileDisconnected = true;
  MQTTAsync_createWithOptions(&mqtt, "tcp://localhost:1883", client_id.c_str(), MQTTCLIENT_PERSISTENCE_NONE, nullptr, &create_options);

  MQTTAsync_connectOptions connect_options = MQTTAsync_connectOptions_initializer;
  connect_options.automaticReconnect = true;
  connect_options.cleansession = true;
  MQTTAsync_connect(mqtt, &connect_options);

  return mqtt;
}

static void SendMessage(MQTTAsync mqtt, std::string topic, std::string payload) {
  MQTTAsync_responseOptions response_options = MQTTAsync_responseOptions_initializer;
  MQTTAsync_send(mqtt, topic.c_str(), payload.size(), &payload[0], 0, false, &response_options);
}

static void ThreadExecute(MQTTAsync mqtt) {
  MQTTAsync_responseOptions response_options = MQTTAsync_responseOptions_initializer;
  while (!terminated) {
    //std::lock_guard<std::mutex> lock(mutex);
    SendMessage(mqtt, "topic", "payload");
    std::this_thread::yield();
  }
}

int main(int argc, char * argv[]) {
  MQTTAsync mqtt1 = CreateClient("mqtt-stress-test1");
  MQTTAsync mqtt2 = CreateClient("mqtt-stress-test2");

  std::thread thread1(&ThreadExecute, mqtt1);
  std::thread thread2(&ThreadExecute, mqtt2);

  for (;;)
    std::this_thread::sleep_for(std::chrono::milliseconds(100));

  return 0;
}

```

Stack trace:
```
Thread #5 [mqtt-stress-tes] 9613 [core: 7] (Suspended : Signal : SIGSEGV:Segmentation fault)	
[...]

PR #689, opened 2019-07-25
Fix issue #572

aggresss:
Fix issue #572 
Segmentation fault when sending a lot of messages from two threads

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -936,6 +936,7 @@ static int MQTTAsync_restoreCommands(MQTTAsyncs* client)
 	FUNC_ENTRY;
 	if (c->persistence && (rc = c->persistence->pkeys(c->phandle, &msgkeys, &nkeys)) == 0)
 	{
+	    MQTTAsync_lock_mutex(mqttcommand_mutex);
 		while (rc == 0 && i < nkeys)
 		{
 			char *buffer = NULL;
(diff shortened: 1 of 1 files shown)


aggresss:
add mutex lock for commands on these functions:
1. MQTTAsync_restoreCommands()
2. MQTTAsync_removeResponsesAndCommands()
3. MQTTAsync_assignMsgId()
4. MQTTAsync_countBufferedMessages()
5. MQTTAsync_getPendingTokens() 
6. MQTTAsync_isComplete()

icraggs:
Hi.  As you can see, this change causes the existing CI tests to fail.  

Also you'll need to sign the Eclipse ECA and each commit as outlined in the PULL_REQUEST_TEMPLATE.md to allow the IP check to pass.

icraggs (commented):
Issue #572 should be fixed by locking the command mutex in countBufferedMessages, nothing more.

icraggs (commented):
Thanks for signing the ECA and the commits.  When fixing an issue, I prefer to focus on the minimum fix needed rather than throwing in other changes too.   The other changes might be good, but also risk unintended side effects as appear to have happened here, as the CI tests are all failing.

icraggs (changes requested):
CI tests need to work.

icraggs:
I've added a fix for #572 specifically, for which the CI tests work.  If any other locking is needed you could put those into another PR, but the CI tests must work.  Thanks.

Issue #701, opened 2019-08-08
Memory leak Definitely when connect failure to server


aggresss:
*Memory leak Definitely when connect failure to server.*
- when conack packet rc not success, receive thread not free conack packet.

```shell
==60== HEAP SUMMARY:
==60==     in use at exit: 720 bytes in 18 blocks
==60==   total heap usage: 562 allocs, 544 frees, 1,359,633 bytes allocated
==60== 
==60== 720 bytes in 18 blocks are definitely lost in loss record 1 of 1
==60==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==60==    by 0x1A81B6: mymalloc (in /mnt/workspace-scratch/linking-device-sdk.v2/build/test/tests/unit/unit_test_qnlink-mqtt)
==60==    by 0x1B97B4: MQTTPacket_connack (in /mnt/workspace-scratch/linking-device-sdk.v2/build/test/tests/unit/unit_test_qnlink-mqtt)
==60==    by 0x1B76E2: MQTTPacket_Factory (in /mnt/workspace-scratch/linking-device-sdk.v2/build/test/tests/unit/unit_test_qnlink-mqtt)
==60==    by 0x1B4B83: MQTTAsync_cycle (in /mnt/workspace-scratch/linking-device-sdk.v2/build/test/tests/unit/unit_test_qnlink-mqtt)
==60==    by 0x1AFE1A: MQTTAsync_receiveThread (in /mnt/workspace-scratch/linking-device-sdk.v2/build/test/tests/unit/unit_test_qnlink-mqtt)
==60==    by 0x4E436DA: start_thread (pthread_create.c:463)
==60==    by 0x517C88E: clone (clone.S:95)
==60== 
==60== LEAK SUMMARY:
==60==    definitely lost: 720 bytes in 18 blocks
==60==    indirectly lost: 0 bytes in 0 blocks
==60==      possibly lost: 0 bytes in 0 blocks
==60==    still reachable: 0 bytes in 0 blocks
==60==         suppressed: 0 bytes in 0 blocks
==60== 
==60== For counts of detected and suppressed errors, rerun with: -v
==60== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```

aggresss:
Pull request https://github.com/eclipse/paho.mqtt.c/pull/702 may be fix it.

PR #702, opened 2019-08-08
Fix memory leak when connect ack rc not success

aggresss:
Fix issue #701 (memory leak when connect ack rc not success)


icraggs:

PR #708, opened 2019-08-14
Reduced, by 1 time, a list search for the MQTTAsync object by socket fd

felixkylo82:


icraggs:
Thanks for the suggestion.  To get a PR accepted straightforwardly, you must ust sign the Eclipse ECA, and git signed-off-by each commit.

I think there may be an issue with the change under some circumstances as the MacOS tests segfaulted.  I've set them to re-run to see whether the outcome is the same.

icraggs (changes requested):
There are other consequences to these changes around the timing of client structure cleanup which cause the CI tests to fail.  It could be quite involved, and unless the CI tests can be shown to work reliably, I don't want to make these changes.  Thanks for your interest.

PR #710, opened 2019-08-14
Fix debian package

klaussnd:
When generating a DEB package, some files were not installed into the correct directories as specified by the Debian standard. For example, `README.md` was installed in `/usr` instead of in the dedicated folder for `libpaho-mqtt`. This PR fixes the location of those files.

Signed-off-by: Klaus Schneider-Zapp <klaus_snd@web.de>


klaussnd:
The failures don't seem to be connected to my changes:
```
40 - test45-7-pending-tokens (Timeout)
```
I have only moved files in the installation step, how can this lead to timeouts?

miks:
Any progress on this?

klaussnd:
I am waiting for the Paho team to look at the PR.

icraggs:
I'm trying to work out how this affects the non-Debian install.  

1. The changes proposed mean that lines 41 to 44 of doc/CMakeLists.txt now become: 

    IF (PAHO_ENABLE_CPACK)
      INSTALL(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/doc/ DESTINATION ${CMAKE_INSTALL_DOCDIR})
    ELSE()
      INSTALL(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/doc DESTINATION ${CMAKE_INSTALL_DOCDIR})

which mean the if statement looks like it's redundant.

2. Is CMAKE_INSTALL_DOCDIR set to a suitable default if PAHO_BUILD_DEB_PACKAGE is not set?

klaussnd:
1. There is a subtile difference between the two branches of the `if`: In the first one, the directory ends with a `/`, which triggers a slightly different behaviour according to the documentation of `install(DIRECTORY ...)`: "The last component of each directory name is appended to the destination directory but a trailing slash may be used to avoid this because it leaves the last component empty." In order to not change existing behaviour, I left the second branch as it was.
2. `CMAKE_INSTALL_DOCDIR` is initialised to `${CMAKE_INSTALL_DATAROOTDIR}/doc/${PROJECT_NAME}` by `GNUInstallDirs.cmake` which is included in the main `CMakefile.txt`.

icraggs:
Ok, well that's a pretty non-obvious distinction (/ or no /).  The test failures are caused by other behaviour changes, not these.  I'm working on those.

klaussnd:
Do you have any suggestion as how to make it more obvious? Should I add a comment?

icraggs:
A comment would be good.  How do I try this out?  This has all been added by contributors, so I haven't exercised it myself.  If it works, instructions could be in the README.

klaussnd:
You can test by generating a Debian package as described in the README. 
I wrote the code such that behaviour of other packages is kept unchanged, but I am not sure that this is actually the correct behaviour. I think that even a `make install` should install the `README.md` of paho into `/usr`.

icraggs:
The CMake build that I tried out ended with 

debian/CMakeFiles/dhtarget.dir/build.make:60: recipe for target 'debian/dhtimestamp' failed

These changes did not affect that result so I'm happy to merge.  At some point I may figure out how the debian build works properly.

Thanks.

klaussnd:
After I created the PR, an additional way to create Debian packages has been added by adding a `debian/` subfolder. This conflicts with the previous way of generating Debian packages by using `cpack`. This leads to the error you mentioned above, and therefore, the original way using `cpack` (which is described in the README) is now broken.
If you remove the `debian/` subfolder and the `ADD_SUBDIRECTORY(debian)` in the main `CMakeLists.txt`, then it works again and you can run:
```
cmake .. -DCMAKE_BUILD_TYPE=Release -DPAHO_WITH_SSL=ON -DPAHO_BUILD_DEB_PACKAGE=ON
make
cpack
```

PR #712, opened 2019-08-15
Fix free invalid pointer in WebSockets

aggresss:
Signed-off-by: Jagger Yu <aggresss@163.com>


icraggs:
I expect to do some work on the websockets function and I will look at this then.  I'll need to think about this more and ideally add a test.

icraggs:
Thanks.  I've had to refactor the WebSocket_putdatas function now so that it is simpler, as a result of fixing the handling of WebSocket frames.  So this PR is no longer needed.

PR #731, opened 2019-09-16
1. Normalize some code;

Doawen:
2.Add comments for some global variables and structure;
3. Add mutex for routine "MQTTAsync_countBufferedMessages", refer to routine "MQTTAsync_assignMsgId";
fix bug: Detecting its elements when the response pointer is null may cause segment fault in 'MQTTAsync_subscribeMany' routine.
fix bug: Allocate room for MQTTProperties but sometimes would forget to free this room when receive PUBLISH response from server.
fix bug: Forget to set topicLen to 0 when strlen(topicName) = topicLen.
fix bug: In receiveThread, forget to free the node's topic string and msg room when user do not set message arrived callback routine.

### DIFF
--- a/src/Clients.h
+++ b/src/Clients.h
@@ -126,7 +126,8 @@ typedef struct
 	willMessages* will;
 	List* inboundMsgs;
 	List* outboundMsgs;				/**< in flight */
-	List* messageQueue;
+	List* messageQueue; /* List queue for PUBLISH message from server that RecevieThread can't deliver immediately when
+	                       connected flag is 0 or deliver failed or other PUBLISH message waiting in queue. */
 	unsigned int qentry_seqno;
 	void* phandle;  /* the persistence handle */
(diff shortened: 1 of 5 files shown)


icraggs (changes requested):
I think it would be better to have these changes in smaller PRs each with focussed changes.  Although the changes to LinkedList do correct formatting, they're not actually needed (but as a separate PR it could make more sense).

The last change, to MQTTProtocolClient.c ping processing, I don't think is correct.  It needs more explanation if you think it is.

icraggs (changes requested):
Also you've change the value of topicLen passed to messageArrived.  Whatever the potential benefits, this is a change to the API behaviour which we can't make right now.

Doawen:
Thankyou for suggestion. I will do the small change and give the distinct comments for every request.

PR #737, opened 2019-10-01
Fix build under Windows

imi0:
Without that define, windows.h will drag a lot of other headers,
including <winsock.h> which is incompatible with <winsock2.h>
used by Paho.

(PS: This seem to only actually trigger when you build with OPENSSL,
as then Stacktrace.c will include winsock2.h and Thread.h at the
same time. Even building without SSL, its never a good idea to have
two headers of different library versions included anywhere in the
project)

### DIFF
--- a/src/Thread.h
+++ b/src/Thread.h
@@ -24,6 +24,9 @@
 #include "mutex_type.h" /* Needed for mutex_type */
 
 #if defined(WIN32) || defined(WIN64)
+    #ifndef WIN32_LEAN_AND_MEAN
+        #define WIN32_LEAN_AND_MEAN
+    #endif
 	#include <windows.h>
 	#define thread_type HANDLE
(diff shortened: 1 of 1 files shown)


imi0:
Mh, not sure how to fix the "eclipsefdn/eca" thingie, when I click on "Details", it takes me to a login page and I neither have an account nor can I create one.

But I hereby release my changes under the same license as Paho is and gift all my "hard work" to you. Is that enough? ^^

(About the continuous-integration/appveyor/pr failing.. the test looks flaky to me. Seems to be an issue for many of the pull requests..)

icraggs:
This option is set as a compile flag in the CMake build for Windows, that's why it's not in the source file.  I think that's the best place?

imi0:
you are completely right :)

PR #739, opened 2019-10-05
Initialize headerlen

### LINKED ISSUE #718: Uninitialized memory causes connection failure
`SocketBuffer.c:207`
headerlen can be uninitialized and very large at this point.
Discovered by _valgrind_;
```
==8165== Conditional jump or move depends on uninitialised value(s)
==8165==    at 0x290C39: SocketBuffer_getQueuedChar (SocketBuffer.c:207)
==8165==    by 0x28D4CE: SSLSocket_getch (SSLSocket.c:810)
```
This causes the connect to fail initially, but it eventually succeeds after a 30 second delay (which I think happens on the broker).

I did this ugly hack to confirm the problem:
`if(queue->headerlen) > 123456789) queue->headerlen = 0;`
which makes the issue go away.

### LINKED ISSUE #740: ssl connection error, probably because def_queue->headerlen is uninitialized memory
I got strange issues when trying to connect to some servers via SSL. It seems to work in release mode, but fails when compiling with cl.exe under Debug.

After some hours of investigation, I think I nailed the issue:

The function SocketBuffer_newDefQ does not initialize the def_queue->headerlen field (it zeroes all the other fields, but not this one..)
Later, the def_queue is inserted into queues with the bogus headerlen and later again, the SocketBuffer_getQueuedChar checks for "if (queue->index < queue->headerlen)"

Since the MS compiler reliably initializes all memory returned by malloc with 0xCDCDCDCD, this headerlen was bigger than the index and it jumped into the if-clause.

Fix for me: Just change line in SocketBuffer_newDefQ from

`def_queue->buflen = def_queue->datalen = 0;`

into 

`def_queue->buflen = def_queue->datalen = def_queue->headerlen = 0;`

tbeu:
Fixes #718.
Fixes #740.

### DIFF
--- a/src/SocketBuffer.c
+++ b/src/SocketBuffer.c
@@ -82,7 +82,7 @@ void SocketBuffer_newDefQ(void)
 	def_queue->buflen = 1000;
 	def_queue->buf = malloc(def_queue->buflen);
 	def_queue->socket = def_queue->index = 0;
-	def_queue->buflen = def_queue->datalen = 0;
+	def_queue->buflen = def_queue->datalen = def_queue->headerlen = 0;
 }
 
 


icraggs:
Thanks.

PR #750, opened 2019-10-15
String Null Checks in UFT-8 validate functions

dpostorivo:
This provides NULL checks for the validate functions. I discovered this while using paho and switching our internal system queuing system sending. Instead of crashing the whole app the passed string is null, it just returns error and the app can handle it.

### DIFF
--- a/src/utf-8.c
+++ b/src/utf-8.c
@@ -80,6 +80,9 @@ static const char* UTF8_char_validate(int len, const char* data)
 	int i, j;
 	const char *rc = NULL;
 
+	if (data == NULL)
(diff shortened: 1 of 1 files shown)


icraggs:
Is this to fix an issue in the Paho library itself, or because you are using these functions elsewhere?

dpostorivo:
@icraggs it's to fix paho itself. Specifically I found this when refactoring our internal code that uses libpaho. We passed paho a null destination in the send message function and it crashed which while it was an error on our part, should not crash the whole application.

icraggs:
Ok.  Thanks.

PR #751, opened 2019-10-18
Fix websocket frame buffer appending

rafjmac:
Without that check there is a possibility to call malloc/realloc with 0 size. The behavior depends on c standard (or library implementation), but most likely:
- malloc will return some pointer (but shall not be dereferenced),
- realloc will free memory (and return NULL),

which shall not happen.

In the scenario when two consecutive calls to `SSLSocket_getdata` are interrupted by the SSL (with `SSL_ERROR_WANT_READ/WRITE`, no data received) and `frame_buffer == NULL`, the following will happen:
- `frame_buffer` will be allocated with `malloc(0)` and it will have valid pointer address (index, len and data_len = 0),
- in the next iteration of the cycle, resize buffer happens and `realloc(0)` is called - `frame_buffer` is `NULL` at that point and is later returned from the`WebSocket_getRawSocketData`.

At the end of the day user experience disconnect.

### DIFF
--- a/src/WebSocket.c
+++ b/src/WebSocket.c
@@ -654,7 +654,7 @@ char *WebSocket_getRawSocketData(
 		frame_buffer = NULL;
 	}
 	// append data to the buffer
-	else if (rv != NULL)
+	else if (rv != NULL && *actual_len != 0U)
 	{
 		// no buffer allocated
 		if (!frame_buffer)


icraggs:
Any chance you have a test which shows how the fix works?  There are other changes needed in the Websockets code too so this would be part of those.

rafjmac:
Unfortunately I haven't written any test due to lack of time - I experienced disconnects on windows when payload was very big (the problem didn't occur on linux - `SSL_ERROR_WANT_READ_WRITE` never occured twice in a row on my linux machine) and decided to fix it in the simplest way - I have been following discussion about websocket implementation and I am aware that this code need redesign. I can imagine writing unit test for this fix but still I am not sure if it is going to be meaningful due to the fact that the behaviour highly depends on libc implementation - I can of course mock calls to `malloc/realloc` but I have no experience in this and don't know how long would it take.

icraggs:
Ok, it's a simple enough change.  More fixes will be needed to the websockets code anyway in due course.  Thanks.

PR #753, opened 2019-10-31
Store initial errno and use the stored one in Socket_error to avoid it being overriden

momobel:
Using MINIMUM logging causes the connection to the broker to fail with musl in alpine docker (even with local broker). PROTOCOL level worked fine. "Invalid argument(22)" error was displayed by Socket_error("connect").

Socket_new calls Socket_error("connect", *sock) if connect fails or is in progress.
Socket_error may call Log directly or though StackTrace logging and any function called may override errno.
For example, Log will call strftime at some point which will call strtoul (with musl).
This causes errno to be set to EINVAL overriding the original EINPROGRESS errno. Caller Socket_new closes the socket.

Avoid this by storing the initial errno before usage in Socket_error.

### DIFF
--- a/src/Socket.c
+++ b/src/Socket.c
@@ -100,19 +100,19 @@ int Socket_setnonblocking(int sock)
  */
 int Socket_error(char* aString, int sock)
 {
-#if defined(WIN32) || defined(WIN64)
-	int errno;
-#endif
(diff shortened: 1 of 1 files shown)


icraggs:
Thanks

Issue #327, opened 2017-09-08
Add pkg-config support


yegorich:
Distribute mqtt-c.pc file providing package version and compiler/linker flags. See https://www.freedesktop.org/wiki/Software/pkg-config/ for more information.

icraggs:
Hopefully PR #756 will resolve this.

PR #756, opened 2019-11-05
Add pkg-config support

GiovanniGrieco:
[pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) is a useful tool to setup the correct compiler flags when needed, especially when setting up linking from another project.

This patch permits the generation of `.pc` files for the compiled libraries. Such files are then installed to the correct `pkgconfig` directory of a Unix-based OS.  Altough pkg-config is also available on Windows and macOS, I do not have enough experience on implementing this feature on that platforms.
Contributions for the enhancement of this PR are welcome.

This PR addresses issue #327.


icraggs:
As it stands, this causes the Linux Travis-CI build to fail (see above).  Some information on how you would use it could be useful, if that makes sense.

icraggs (changes requested):
The Linux build on Travis fails with:

CMake Error at /home/travis/build/eclipse/paho.mqtt.c/build.paho/cmake_install.cmake:49 (file):
  file cannot create directory: /usr/local/lib/pkgconfig.  Maybe need administrative privileges.

We shouldn't use such directories unless we're doing an install, I guess, when you would expect to have admin privileges.

Issue #565, opened 2018-10-15
windows static build MQTTClient_subscribe Access violation reading


jumoog:
If i use the static lib sometimes MQTTClient_subscribe throws a Unhandled exception..  Access violation reading.

Its random and doesn't happen with the none static lib. I guess its socket releated.

No idea how to debug this issue

OS: Windows

icraggs:
I've never tested or used the static library.  It was added as an option by a contributor, but without tests.  It has seemed to work for people who have used it, but really there should be tests.

Options for debugging:
- set the client library trace environment variables (see readme) and let's see if the trace output helps
(- run under Visual Studio and get detailed execution info
- run on Linux or MacOS under gdb/lldb and get a backtrace)

jumoog:
(-2)> MQTTClient_subscribe5:1945
(-1)> MQTTClient_subscribeMany5:1813

but i set MQTTVERSION_3_1_1

icraggs:
Those calls are still used internally whichever version of MQTT you are using.  Do you have a snippet of the code where you call subscribe?

jumoog:
here i made an example: 
sometimes it connects fine and sometimes it crashes
https://gist.github.com/jumoog/3cdaf968f622e7f7af63c8e9935171f8

if i use the none static libs i can connect every time

icraggs:
That code look ok.  Does the trace show the same output each time?  Can you attach a fuller trace as a file?  If you don't subscribe does everything work?

jumoog:
it took forever to find a crash without subscribe ~30 times

###with subscribe
[error.txt](https://github.com/eclipse/paho.mqtt.c/files/2496864/error.txt)
[error2.txt](https://github.com/eclipse/paho.mqtt.c/files/2496867/error2.txt)
###without subscribe
[crashWithOutSub.txt](https://github.com/eclipse/paho.mqtt.c/files/2496881/crashWithOutSub.txt)

jumoog:
on my Server it works perfect!

the same binary keeps crashing on my home network. I guess its Socket related.

[Home.txt](https://github.com/eclipse/paho.mqtt.c/files/2498489/Home.txt)
[Server.txt](https://github.com/eclipse/paho.mqtt.c/files/2498493/Server.txt)

icraggs:
What is the hardware/OS you are using at home?  That home.txt trace doesn't look right at all.  Can you try using dynamic library option there to see if it makes a difference?

[...]

PR #761, opened 2019-11-20
Initializing static library on Windows

uhuntu:

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -2809,6 +2809,24 @@ int MQTTAsync_connect(MQTTAsync handle, const MQTTAsync_connectOptions* options)
 	int rc = MQTTASYNC_SUCCESS;
 	MQTTAsync_queuedCommand* conn;
 
+#if defined(WIN32) || defined(WIN64)
+	if (mqttasync_mutex == NULL)
+	{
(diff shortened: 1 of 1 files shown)


icraggs:
I'm not sure what problem you're trying to solve here?

uhuntu:
> I'm not sure what problem you're trying to solve here?

When I am trying to use statically build on windows, I met the trouble.

jumoog:
i guess its related to #565 
https://github.com/eclipse/paho.mqtt.c/issues/565#issuecomment-463221351

icraggs (changes requested):
Ok.  There are two problems with this:

1) the code is duplicated from DllMain - if it's going to be reused it should be put into a function and then called from wherever it's needed.

2) By the time we get to connect(), mutexes have already been used - it's too late.  Create() or global_init is where initialization is currently done.  The create() call also uses a mutex at the top, so that won't work either, leaving two options: 
a) a static initialization (which may have to be conditional on a static build)
b) in global_init() which will then have to be called when using the library statically.  
Using global_init() will provide most control to the application, I think.

icraggs:
Ok, this is fixed in PR #341, which I've now merged into the develop branch.  Thanks.

PR #780, opened 2019-12-30
only set OPENSSL_SEARCH_PATH if not empty

waltronix:
Fix problem with always overwriting the open sll search path to the default value.

Signed-off-by: Andreas Walter <andreas.walter@waltronix.de>


icraggs:
Thanks.

Issue #721, opened 2019-09-03
Large Payloads over WSS either disconnects or fails PUBACK


keysight-daryl:
Using paho c++/c 1.3.1 (windows) via async client over WSS to erlio/docker-vernemq:1.9.1 broker we publish messages with payloads > 4000 bytes and subscribe to the same topic.  In a loop on the client we validate arrival of published messages and that the payload of what is received matches what was sent. This works fine for payloads under 4000 bytes however for payloads above this size we have noticed consistent failure of the client to PUBACK and at times being disconnected from the broker.  We do not publish the next message until we have validated receipt of the previous published message subscription and so in theory there should only ever be one message in flight.  Switching to a connection over TCP does not have this issue.  Point of failure is somewhat random in nature but easy to reproduce and will occur within the first few messages for payloads over 4050 bytes and never for payloads under 4000 bytes. Between 4000-4050 bytes it occurs somewhat randomly.  Attached is a broker log of one such failure where the client appears to not have performed the PUBACK.

![mqttIssue](https://user-images.githubusercontent.com/40178586/64202435-20433080-ce5f-11e9-8c7a-e9d37edb0789.PNG)

icraggs:
Do you know, or can you try whether you get the same results against Mosquitto, or the test server in https://github.com/eclipse/paho.mqtt.testing?

keysight-daryl:
our application uses WSS and some custom auth so it might not port so easily around.  We have noticed the issue is less pronounced using WS without SSL requiring larger payloads (around 8500 bytes) to cause trouble.  We have tried to duplicate the issue using the paho.js client without success so it points to this being a bug within paho.c/cpp.  I will see if I can target the test server you mentioned.

icraggs:
Alternatively, if it's easy to reproduce, taking a client library trace of the problem at the default level, as described in the readme, could help.

keysight-daryl:
So same issue occurs against wss://test.mosquitto.org:8081/mqtt.  It works fine for small payloads like 1k bytes but hangs on the ack and never receives the subscribed message for larger (5k) payload sizes.  Attached a screen grab of it failing on that broker...  Which readme describes grabbing a debug trace?  I'll try that next.

![mqttIssueMosquitto](https://user-images.githubusercontent.com/40178586/64292405-56081800-cf38-11e9-996e-2ebf43d9fbea.PNG)

keysight-daryl:
[...]

PR #787, opened 2020-01-03
Large Payloads over WSS either disconnects or fails PUBACK #721

keysight-daryl:
Resolves issue #721 where client was forcing a disconnect as a result of failing to receive the first 2 bytes of a websocket frame during a websocket peer renegotiation in which OpenSSL::SSL_read(...) is expected to return SSL_ERROR_WANT_READ until the renegotiation is complete.  Also addresses a bug in getRawSocketData() not using the caller bytes requested to fulfill the request and avoid use of a c++ language keyword 'final'.

Signed-off-by: Daryl Fortney <daryl.fortney@keysight.com>

### DIFF
--- a/src/WebSocket.c
+++ b/src/WebSocket.c
@@ -664,6 +664,8 @@ char *WebSocket_getRawSocketData(
 {
 	char *rv;
 
+	size_t bytes_requested = bytes;
+
 	FUNC_ENTRY;
 	if (bytes > 0)
 	{
@@ -735,6 +737,8 @@ char *WebSocket_getRawSocketData(
 	else 
 		goto exit;
 
+	bytes = bytes_requested;
+    
 	// if possible, return data from the buffer
 	if (bytes > 0)
 	{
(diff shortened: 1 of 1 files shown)


icraggs:
The changes look fine, but I've not found a scenario (for me) which is affected by them yet.  Receiving a 'large' packet is still failing.  Large being > 8k or thereabouts, which is obvs not that large.

keysight-daryl:
unit tests seem to be failing that are a result of a previous pr commit not this one?

keysight-daryl:
interesting it is working great for us with hundreds of messages over 10k... one difference however is we are still using the last stable release without all the other changes in this branch.  not sure if some other change might be causing the issue you are seeing.

icraggs:
Ok, that's great that it's working for you!  Thanks for the PR.  Yes the CI tests are failing due to the MacOS Travis environment having problems recently - it's really very slow.

PR #796, opened 2020-01-10
define library version via version files

waltronix:
Extract versioning via files from @rpoisel pull request as suggested by @icraggs in https://github.com/eclipse/paho.mqtt.c/pull/181#pullrequestreview-333485209.

Signed-off-by: Andreas Walter <andreas.walter@waltronix.de>


icraggs:
Looks good.  Thanks.

PR #798, opened 2020-01-17
Reduce symbol visibility in cmake linux builds

### LINKED ISSUE #794: internal symbols are exported on linux
On linux, the shared libraries do not just export symbols which are supposed to be exported but also tons of internal symbols, some of which can very easily conflict with other libraries or user programs, e.g. "state", "writeInt", "Log", and "SHA1_Update".

Please ensure that code is compiled with -fvisbility=hidden and add suitable attributes to declarations of symbols that need to be exported. See https://gcc.gnu.org/wiki/Visibility for details.

waltronix:
Fixes #794 

Unified export declarations and fixed visibility for cmake linux builds.
 
I will open a new issue that picks up our discussion on updating cmake builds.


src/ExportDeclarations.h
```diff
@@ -0,0 +1,36 @@
+/*******************************************************************************
```
waltronix, line 1: Is a header required? I put it there, because the other files also had it.

waltronix:
@icraggs I would be happy, if you find time to have a look on this. After it is merged, I would continue with #799

icraggs:
Thanks, it looks good.

PR #800, opened 2020-01-19
In Websocket mode, make best effort to retrieve the right amount of data

frbouvi:
This pull request fix a problem where already fetched incomplete data can block the whole frame exchange in websocket.

If a frame is already there but with not enough data, fetch the remaining bytes from the socket, then merge everything in a single buffer.


icraggs:
@frbouvi so this is obsolete now?

icraggs:
You said disregard, so I will close.

frbouvi:
Yes it is

-Frédéric Bouvier

________________________________

PR #801, opened 2020-01-21
Fix the last websocket fix by ensuring pointers are updated after a realloc

frbouvi:
Makes the last fix works for me


icraggs:
That's great, thanks.  I tested the changes on Linux and they worked there, by accident evidently - not so on Windows.

Issue #792, opened 2020-01-06
Static Linkage


fpagliughi:
I'd like to get static linking as a first-class citizen for this project. Over the last few years, I've had several customers requesting static linking for their embedded Linux projects. I suppose now that flash space is so cheap, dynamic libraries are not as important as they used to be.

Plus, the Paho Rust client uses static linking by default, and thus is completely broken on Windows. It doesn't give any warnings; it builds fine, but then all the apps crash hard due to problems with mutexes. So this is holding up the next (MQTT v5) Rust release.

In the long run, I'd like to accomplish two things:
- Make sure the static library works on all supported platforms
- [Less important] Update CMake so that it can build just the static library and not the dynamic one, on request.

This issue appears to be scattered across a number of open issues: #261, #263, #398, #519, #655, #660, etc...

If any work has started in this area, let me know, otherwise I can start working on it right away.

jumoog:
maybe its better to write it completely in rust add we can use it via FFI?

icraggs:
@fpagliughi The develop branch has some changes which I think should allow static linking to now work on Windows.  It would be good if you could try that out.

fpagliughi:
@jumoog I thought of that. Perhaps as the async/await stuff in Rust stabilizes a bit more, it would be worth considering a 100% Rust library. 

Then, again, I thought v2 of the C++ library would be 100% stand-alone C++. And that was +6yrs ago...  So, we'll see.

@icraggs Cool! I'll try it out this week.

fpagliughi:
The static build seems broken even on Linux at the moment. I get this:
```
...
[ 51%] Building C object src/CMakeFiles/common_obj.dir/Base64.c.o
[ 52%] Building C object src/CMakeFiles/common_obj.dir/SHA1.c.o
[ 53%] Building C object src/CMakeFiles/common_obj.dir/WebSocket.c.o
[ 53%] Built target common_obj
Scanning dependencies of target paho-mqtt3c-static
[ 54%] Building C object src/CMakeFiles/paho-mqtt3c-static.dir/MQTTClient.c.o
/home/fmp/mqtt/paho.mqtt.c/src/MQTTClient.c:77:10: fatal error: VersionInfo.h: No such file or directory
 #include "VersionInfo.h"
          ^~~~~~~~~~~~~~~
compilation terminated.
[...]

PR #803, opened 2020-01-23
Fix static build.

zod:
This should at least fix the linux build errors mention in #792


icraggs:
Thanks.  To merge automatically the IP checks would have to pass - I've added the changes.

PR #805, opened 2020-01-24
Make the shared library build optional in cmake.

deonmcclung:
Make the shared library build optional for cmake generators that do
not support the creation of shared libraries, as is the case with some
embedded architectures.


icraggs:
This is ok except for the changes to MQTTAsync.h which I'm not keen on.  A recent PR has moved those declarations into a separate file, ExportDeclations.h (in the develop branch).  So maybe there's a way to conditionally choose one or the other depending on the current library being built.

fpagliughi:
Yes, I agree with @icraggs that this should be kept in a separate file. In fact, I believe that CMake can automatically generate export headers for the platform and target, although I've never tried it:
https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html

Can @waltronix shed some light on this?

Also, if this is in MQTTAsync.h, is it also handled for the synchronous version of the library?

fpagliughi:
BTW, I'm sure I recently watched a video explaining this and some stuff from #798, in case anyone wants to dive into background. I think this was it:
https://www.youtube.com/watch?v=m0DwB4OvDXk

deonmcclung:
Yeah, I don't disagree. I hesitated to make the change in src/MQTTAsync.h. I am open to suggestions on that. I'm trying to think of how even CMake can help make a situation where a single header file can support two different configurations of the library at the same time. When a client source file includes MQTTAsync.h, is the compiler to be expected to generate code to pull in functions from an import library --- __declspec(dllimport) --- or is it not?

This is a situation where we are trying to support generating two versions of the library at the same time. It is a legacy situation that I tried to preserve, but I can't imagine that condition is actually useful to anyone.

I could easily address this if I made it so the library built as static OR shared, but not both at the same time. Would anyone be opposed to this solution?

Thanks for looking at this.

icraggs:
Personally, I think that having either a static or shared build on each invocation of CMake is fine.   Please sync up with the latest changes in the develop branch as well.

waltronix:
In other projects, I have used manually export headers with an additional 'STATIC_LIB' define to provide an empty version of the export declaration. That would be fine I think, but we would loose the ability to easily reuse the generated object files for static and dynamic libraries. This will cost some extra compilation time, but would make the cmake file a bit simpler.
@fpagliughi The cmake generated header stuff looks nice, I have to have a deeper look on it.

src/CMakeLists.txt
```diff
@@ -120,10 +124,22 @@ IF (PAHO_BUILD_STATIC)
                 ${CMAKE_BINARY_DIR})
     ENDFOREACH()
 
-    INSTALL(TARGETS paho-mqtt3c-static paho-mqtt3a-static
-        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+    IF (NOT PAHO_BUILD_SHARED)
```
waltronix, line 127: Is it correct, that static libraries will only be installed, if no shared libraries are build?

deonmcclung:
> Personally, I think that having either a static or shared build on each invocation of CMake is fine. Please sync up with the latest changes in the develop branch as well.

So, I essentially started over from the latest on the develop branch, going forward with the assumption that if the static build is active, the shared build is not. It actually simplified things quite a bit.

I appreciate the suggestions given. I used GenerateExportHeader the existing export filename and defines.

src/CMakeLists.txt
```diff
@@ -69,12 +69,23 @@ ENDIF()
...
+)
 
 ADD_EXECUTABLE(MQTTVersion MQTTVersion.c)
+TARGET_INCLUDE_DIRECTORIES(MQTTVersion
+    PRIVATE ${CMAKE_BINARY_DIR}
+)
 
-ADD_LIBRARY(paho-mqtt3c SHARED $<TARGET_OBJECTS:common_obj> MQTTClient.c)
-ADD_LIBRARY(paho-mqtt3a SHARED $<TARGET_OBJECTS:common_obj> MQTTAsync.c)
+IF (PAHO_BUILD_STATIC)
```
waltronix: I think you removed your definition of the `PAHO_BUILD_STATIC` cache variable.

src/CMakeLists.txt
```diff
@@ -85,6 +96,13 @@ SET_TARGET_PROPERTIES(
     VERSION ${CLIENT_VERSION}
     SOVERSION ${PAHO_VERSION_MAJOR}
     DEFINE_SYMBOL "MQTT_EXPORTS")
+
+GENERATE_EXPORT_HEADER(paho-mqtt3a
```
waltronix: If we generate the export declarations header, we should remove the manually written one.
But this will break build via make (@icraggs )

deonmcclung:
Whoops. Yes, you’re right. I had that on my first submission of this re-attempt, but I didn’t have the licensing right on my commits. I missed that when I re-made them. I’ll fix it.

src/CMakeLists.txt
deonmcclung: I don't believe so... I did start over from my first submission back to the develop branch. I'm sorry if that made things confusing.

src/CMakeLists.txt
fpagliughi: Oh, yes, you're right @waltronix. I had forgotten about the make build in this repo.

src/CMakeLists.txt
deonmcclung: I removed the `PAHO_BUILD_SHARED` that I had added. Now, if the previously existing `PAHO_BUILD_STATIC` is set, the library will only build the static libraries _and not the shared_ (new behaviour). If it is not set, the shared will build and not the static -- except for the import libraries on Windows (existing behaviour).

src/CMakeLists.txt
fpagliughi: I would prefer, in general, that it be left up to the user whether to build and install the static libraries, shared libraries, or both. Certainly CMake allows optimization of the build in the case of both by using object libraries so that the sources are only compiled once and then linked into the two libs.

With both static and shared libs installed, it's then up to each application or project to decide which lib is better for the project. (As I juggle multiple projects/customers simultaneously, this is essential).

I will look at this in detail later today, but in the meantime, here is an example of what I did for another library, albeit with my limited CMake skills:
https://github.com/fpagliughi/sockpp/blob/master/CMakeLists.txt

deonmcclung: I believe I understand what you are saying. However, building both libraries at the same time presents a difficulty. At least for windows, it is not simply a link-line choice for the consumer between the static and the shared libraries. To make the _both_ option work, the end consumer must set a `#define` at compile time to specify whether they desire to link against the static or shared version of the library in addition to specifying the correct library on the link-line.

This is because when a header declares the following function:

`int MQTTAsync_reconnect(MQTTAsync handle);`

...after preprocessor macros are evaluated, the code will either evaluate to:

`__declspec(dllimport) int MQTTAsync_reconnect(MQTTAsync handle);`

or 

`int MQTTAsync_reconnect(MQTTAsync handle);`

but it cannot be both at the same time. You need an additional `#define` from the user to specify which one of the two declarations is desired by the user _at compile time_. The first tells the compiler to expect the DLL import library on the link line, the second works for the static library.

The second line (without the __declspec) technically works for both the static and DLL libraries. However, according to Microsoft documentation, specifying the dllimport causes the compiler to generate more efficient code. Just removing the dllimport was the approach I took on my first submission, but that was undesired.

Making it build one way or the other simplifies the CMake files quite a bit. Additionally, there shouldn't actually be any end-user that should actually _need_ both configurations at the same time.

icraggs:
I don't think it's working yet, but we'll see what happens to the CI tests.

If we are going to have one or the other builds, shared or static, then I think that's a good plan.  Then the samples and tests should be linked in the right mode too, and all the tests run.  This means we can have the static and shared builds run consecutively in one CI build, or as separate CI builds, as we see fit.

I did a quick experiment by adding this to the test/CMakeLists.txt file:

IF (PAHO_BUILD_STATIC)
ADD_EXECUTABLE(
	test1-static
	test1.c
)

TARGET_LINK_LIBRARIES(
	test1-static
	paho-mqtt3c-static
)
ENDIF()

and got a test1 executable, statically linked, which worked (on Linux).  

This doesn't necessarily all have to be done in this one PR, though.  I just wanted to be clear what I think we need to be aiming for.

icraggs (changes requested):
On Ubuntu 18.04 with CMake 3.10.2, I'm getting:

> -- Performing Test COMPILER_HAS_HIDDEN_VISIBILITY

> CMake Error at /usr/share/cmake-3.10/Modules/CheckCXXSourceCompiles.cmake:97 (try_compile):

  
> Unknown extension ".cxx" for file
  
>  /home/icraggs/git/paho.mqtt.c/build.paho/CMakeFiles/CMakeTmp/src.cxx

>   try_compile() works only for enabled languages.  Currently these are:

    
> C

> See project() command to enable other languages.

> Call Stack (most recent call first):
>   /usr/share/cmake-3.10/Modules/CheckCXXCompilerFlag.cmake:49 (CHECK_CXX_SOURCE_COMPILES)
>   /usr/share/cmake-3.10/Modules/GenerateExportHeader.cmake:215 (check_cxx_compiler_flag)
> /usr/share/cmake-3.10/Modules/GenerateExportHeader.cmake:374 (_test_compiler_hidden_visibility)
>   src/CMakeLists.txt:100 (GENERATE_EXPORT_HEADER)

> -- Performing Test COMPILER_HAS_HIDDEN_VISIBILITY - Failed

icraggs:
I fixed that by adding a line:

ENABLE_LANGUAGE(CXX)

to the main CMakeLists.txt.

icraggs:
And, how are we going to allow the Makefile to work?  Put the header in another directory and copy it, perhaps, so it doesn't get confused with the generated one?

deonmcclung:
> And, how are we going to allow the Makefile to work? Put the header in another directory and copy it, perhaps, so it doesn't get confused with the generated one?

I'm sorry, @icraggs, I don't follow you on this one. Could you please elaborate?

icraggs:
The plain make build on LinuxMac doesn't work because there is no ExportDeclarations.h?

deonmcclung:
> The plain make build on LinuxMac doesn't work because there is no ExportDeclarations.h?

I've tried it with WSL Ubuntu, and I don't seem to have a trouble building the Linux via make. I don't have access to a full Linux system at the moment. Would that be different?

Just to be clear. In my WSL Linux build, inside my build.paho directory, I see a valid ExportDeclarations.h. Also, after generating the CMake, I ran a make -n, I see (one of the lines):

`cd /home/dmcclung/devel/paho.mqtt.c/build.paho/src && /usr/bin/cc -DMQTT_EXPORTS=1 -D_GNU_SOURCE -I/home/dmcclung/devel/paho.mqtt.c/build.paho  -fPIC   -fvisibility=hidden -o CMakeFiles/common_obj.dir/WebSocket.c.o   -c /home/dmcclung/devel/paho.mqtt.c/src/WebSocket.c`

Most notibly is the:
` -I/home/dmcclung/devel/paho.mqtt.c/build.paho`

...which should find the ExportDeclarations.h.

icraggs:
I meant using the Makefile included in the git repo, not the one generated by CMake.  I don't want that to stop working.  The CMake build is fine.  I'll think about it.

fpagliughi:
Apologies for being slow to respond on this. I've been terribly busy at work the last few weeks.

I _really, really, really_ don't want this to just be an option of either the shared library or the static library.
_Really, really, really_.

First of all, that's not how most libraries work these days, and also because I personally need both. On all targets. And I'm not sure that I'm alone on this. I'm also worried about the Paho C++ build, since it can have both, and links to the respective C lib in either case.

Anyway, it's so common that CMake provides an easy way to do it with object libraries.

The general rules that I follow for CMake these days are:
1. Allow static or shared libraries to be built. Or both.
2. Default to shared ON, static OFF.
3. Build the sources to a CMake object library, then create the static or shared library from it (so that the files are only compiled once even if both libs are built).
4. Use the same base file name for the lib on *nix targets, use `mylib-static` for static libs on Windows.
    (a) On *nix, use the name `mylib.so` for shared libraries, `mylib.a` for static libraries. That's pretty standard.
    (b) On Windows, this isn't possible since they use the `.lib` extension for both static libraries and dynamic import libraries. So name the shared library, `mylib.dll` (with a `mylib.lib` import), and `mylib-static.lib` for the static library.
5. The CMake build needs to explicitly decide which lib to use for the tests and samples if both are created.

deonmcclung:
> I meant using the Makefile included in the git repo, not the one generated by CMake. I don't want that to stop working. The CMake build is fine. I'll think about it.

@icraggs Oh, I didn't realize that build was out there too. I'll see if I can get that one to work also.

deonmcclung:
@fpagliughi I'm very sorry. I am just trying to do what people want. My original posting preserved both configurations. I haven't personally come across any other libraries that permit the build of both types at the same time, and other than as a convenience/optimization for a build-farm, no program should actually try to use both configurations at the same time. So, I don't know, I will do whatever the collective desires.

However, I do still have a question for you. What do you propose about the __declspec(dllimport) issue? (See https://github.com/eclipse/paho.mqtt.c/pull/805#issuecomment-579520548)

To the client application, the header can only, after macros are evaluated, declare a function to be ONE of:

`__declspec(dllimport) int MQTTAsync_reconnect(MQTTAsync handle);`

or

`int MQTTAsync_reconnect(MQTTAsync handle);`

It cannot be both at the same time. The first declaration works with DLLs and according to Microsoft, produces the most efficient code. The second will work against a static library or with the import library, and was how I had it on my first submission, but that was not desired. (See https://github.com/eclipse/paho.mqtt.c/pull/805#issuecomment-579415933)

To build both does not simply require a the end-consumer to make a Link-time choice as to which library to link. They have to have their compile match the link (unless we ditch the dllimport tag all together).

When you build both at the same time, do you propose adding an additional #define to the client program to get it to choose the correct dllimport specification?

I'd be interested in your suggestion.

fpagliughi:
Hey @deonmcclung. Yes, I understand you are getting conflicting requests (sorry @icraggs !).

A single application will link to one library or the other. That is true. But if you have many applications for many different projects that use the Paho C library, some may want static linking, and others may want shared linking. And so it would be preferable to build and install the libraries once, and then let the applications chose what they want.

The Paho C++ library does this. It allows you to build one or the other or both:
https://github.com/eclipse/paho.mqtt.cpp/blob/master/CMakeLists.txt
https://github.com/eclipse/paho.mqtt.cpp/blob/master/src/CMakeLists.txt

The existing C lib in the _develop_ branch allows you to do just shared or both. I would like to add the option of just static without removing either of the other options.

Recently (on the C++ issues) someone with a weird target reported being able to build the static library, but unable to build the shared one. They would have been happy with just the static lib, but there is no way to do that. Thus, this issue popped up.

fpagliughi:
Keep in mind, too, that this library is not just used for creating C programs. It also forms the base for the Paho C++ and Rust libraries. So people with a lot of different requirements are using the library, and having flexibility (without added confusion) is better.

icraggs:
I was asked my opinion about whether building shared and static builds separately was ok, and I just gave my opinion :-)  It wasn't a hard and fast requirement.

If both shared and static libraries are built at the same time, then the tests (and perhaps samples) will need to be built twice two, with non-conflicting names and/or locations.  Which is why I experimented with

IF (PAHO_BUILD_STATIC)
ADD_EXECUTABLE(
test1-static
test1.c
)

TARGET_LINK_LIBRARIES(
test1-static
paho-mqtt3c-static
)
ENDIF()

to see if it works.  It does.

deonmcclung:
@fpagliughi I see your point. My initial thought with my change was to preserve existing expectations on the library. So with that, I guess the only option left is to go back to something close to my initial delivery, where I remove usages of dllimport. To my understanding, it will still work against the DLLs without that. 

I will make a new submission to that end. To be frank, I don’t have a lot of experience making open-source submissions. I don’t know how decisions are made. I just need the shared library build to be optional somehow, and thought that feature might be useful to someone else in a similar circumstance as me.

icraggs:
@deonmcclung thanks for working on this.  There is no one way that decisions are made on open source projects, so there's not need to feel that you're missing something.  On this one, I'm responsible for this C project, but I don't have as much experience actually using it as some other people, like Frank.

Being able to choose to build shared, statically or both is obviously the best of all worlds.   I'm happy to continue working  on this, taking over from you at whatever point you would like (which might be now!)

If we build both static and shared at the same time, then the test (and sample) programs will have to be in different locations and/or have different names.  Especially the tests, so that they can both be run if we have both static and shared builds built.   At the moment, if the static library only is built, then the sample linking will fail. 

So I think in the test CMakeLists.txt for instance we should have:

    IF (PAHO_BUILD_STATIC)
        ADD_EXECUTABLE(
            test1-static
            test1.c
        )
        TARGET_LINK_LIBRARIES(
            test1-static
            paho-mqtt3c-static
        )
    ENDIF()

    IF (PAHO_BUILD_SHARED)
        ADD_EXECUTABLE(
            test1
            test1.c
        )
       TARGET_LINK_LIBRARIES(
            test1
            paho-mqtt3c
        )
    ENDIF()

which will build both static and shared versions of test1, and allow us to run either one or both.  Does that make sense?  As I said I'm happy to take it from here if you want.

deonmcclung:
@icraggs It is telling me that changes are requested, but I'm not exactly sure what the request is. Is there anything else you are looking for?
Thanks!

icraggs (approved):

icraggs:
Those changes were for the original Makefile working - which you have fixed, so that's great.

deonmcclung:
@icraggs Thanks for the words of encouragement. I can make another delivery arranging the tests how you mentioned. That seems straight forward. I do appreciate the guidance on that.

fpagliughi:
Yes. I'm really glad that folks are working on the build system. (Thanks @deonmcclung  & @waltronix ). I wish I could be more help than just giving a wish list, but my CMake skills are not yet up to the task. But I'm trying to learn more, and maybe in the next round I can be a bit more helpful.

@icraggs I hadn't thought to do tests using both versions of the library. In some of my other projects, I just use the shared library if it's defined, but if not, then I use the static library. (It's an arbitrary choice). For example, in my C++ socket library, _sockpp_ I have:
```
# --- Default library for examples and unit tests ---

if(SOCKPP_BUILD_SHARED)
    set(SOCKPP_LIB ${SOCKPP})
else()
    set(SOCKPP_LIB ${SOCKPP_STATIC})
endif()
``` 

Then, for the test executable(s):
```
# --- Link for executables ---

message(STATUS "Using library for unit tests: ${SOCKPP_LIB}")

target_link_libraries(unit_tests ${SOCKPP_LIB} 
	Catch2::Catch2
	Threads::Threads
)
```

deonmcclung:
> At the moment, if the static library only is built, then the sample linking will fail.

I believe I did address this problem already by making it so that if the shared version of the library is not being built, it will try to link against the static. Otherwise, it links against the shared.

Do you still want to make two separate versions of the tests?

icraggs:
I have made changes to the test CMakeLists file which builds both static and shared versions of the test programs, so that they can be both run if needed.  I prefer it like that - no confusion about how which tests are liniked, and the static tests won't get missed accidentally.

So we could go with that.

deonmcclung:
@icraggs Sounds good to me. Are you saying you already have the changes, or have you just verified that they work and they still need to be made? BTW, I made one more push to update the README.md.

icraggs:
I've made the changes locally to see how they work - apart from running all the static tests.  Looks good to me.  So we could merge all this into the develop branch.

Issue #368, opened 2017-11-16
Can't build code with C++, missing underscores in MQTTAsync.h


gkiryaziev:
hi, VS2015 can't build my C++ code, and I found missing underscores here

https://github.com/eclipse/paho.mqtt.c/blob/master/src/MQTTAsync.h#L95

have to be 

`#if defined(_WIN32) || defined(_WIN64)`

and I think MQTTClient.h have this problem too

https://github.com/eclipse/paho.mqtt.c/blob/master/src/MQTTClient.h#L112

VC2015 don't recognize Windows

icraggs:
This code has been building in various versions of Visual Studio for 10 years or more, without those definitions being changed.  I don't think it's the source of your problems.

How are you building the code?  Are you using CMake to generate the Visual Studio project?

icraggs:
I've just built the code unchanged with Visual Studio 2015, using the cbuild.bat script (using CMake).

gkiryaziev:
Hi, I am using CMake for C and C++ libraries and then successfully build them in VS2015 x64, I have no problem with libraries building, I have a problem then when I use the C++ library in my project, because I have to include MQTTAsync.h in the project and then VS2015 does not detect Windows.

here you can see how VS2015 selects `else` section without underscores

![2017-11-18_153207](https://user-images.githubusercontent.com/8873218/32980813-2c691a60-cc76-11e7-86ce-f30d3ff8c614.png)

![2017-11-18_153234](https://user-images.githubusercontent.com/8873218/32980848-a65ffd84-cc76-11e7-9668-64be8f1e7fe0.png)

and finally with underscores my code was built successfully

icraggs:
The CMake build includes test programs and sample applications too.  I'm not sure what you mean by the C++ library - is that the separate Paho C++ client library?

You could just add the WIN32 definition, if it isn't defined already.

gkiryaziev:
I am using [MQTT C++ Client for Posix and Windows](https://www.eclipse.org/paho/clients/cpp/) and it requires the [Paho MQTT C client library](https://www.eclipse.org/paho/clients/c).

Yes, I can add the `WIN32` or `WIN64` definition, before `#include` in my code, like this:

`#ifndef WIN64`
`#define WIN64`
`#endif`

`#include <iostream>`
`#include "mqtt\client.h"`

apparently this is the fastest trick. Thank you.

fpagliughi:
[...]

PR #806, opened 2020-01-25
#368 Add underscores to Windows compiler macro checks, _WIN32 and _WIN64

fpagliughi:
Signed-off-by: fpagliughi <fpagliughi@mindspring.com>

This updates the conditional compilation checks for Windows to use the _WIN32 and _WIN64 pre-defined macros (with the leading underscore).

### DIFF
--- a/src/Base64.c
+++ b/src/Base64.c
@@ -16,7 +16,7 @@
 
 #include "Base64.h"
 
-#if defined(WIN32) || defined(WIN64)
+#if defined(_WIN32) || defined(_WIN64)
 #pragma comment(lib, "crypt32.lib")
 #include <windows.h>
 #include <wincrypt.h>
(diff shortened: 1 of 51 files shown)


icraggs:
Thanks Frank.  My only reservation is affecting very old versions of Visual Studio.

icraggs:
We'll find out I guess :-)

fpagliughi:
Microsoft doesn't even appear to be supporting compilers older than 2015, as far as I can tell from the on-line docs. And this covers back to then (and probably way further back).

If anything does come up, I suppose it wouldn't hurt to make the checks either/or:
```
#if defined(WIN32) || defined(_WIN32) || defined(WIN64) || defined(_WIN64)
```
but I wouldn't bother unless something comes up.

fpagliughi:
Oh, and if we did have to do something silly like that, it might be better to isolate it in a base header somewhere rather than scattering it throughout the code. Have that long #if define a "paho" windows macro, then use that, like:
```
#if defined(WIN32) || defined(_WIN32) || defined(WIN64) || defined(_WIN64)
#define PAHO_WINDOWS
#endif 
```
Then, throughout the code:
```
#if PAHO_WINDOWS
    // Windows code
#else 
    // non-windows code
#endif 
```
But, again, only if there's a problem with the way it is now...

icraggs:
We can do that if the problem arises.  I did have an issue a year or two back where I change I made had broken a user's VS2010 build :-)

Issue #799, opened 2020-01-17
Update cmake build system


waltronix:
As discussed in #794 we should update the cmake build system to support all features of the Makefiles with the goal to replace the Makefiles.

Suggested is an update of the required cmake version to 3.5 to support builds on ubuntu 16.04 LTS.
This would also be beneficial for usage of OpenSSL on windows.

In context, I also suggest to update the CI build environments to use VS2019 with cmake (and maybe clang).

waltronix:
@icrags a general question, would you prefer a series of smaller pull requests or a bigger one with small commits?

jumoog:
but i need VS2013 / VS2015 support 🤕

waltronix:
According to cmake docs, it should be available.

https://cmake.org/cmake/help/v3.5/generator/Visual%20Studio%2012%202013.html

jumoog:
my bad i thought your idea was dropping support for older versions

icraggs:
@waltronix smaller PRs are always good - easier to review and test.

The CI tests should use both recent and old versions of Visual Studio, IMO.  Ideally all of them that are supported, but that's probably too expensive in running time and CI resources.

icraggs:
Visual Studio 2013 and 2015 are used in the AppVeyor CI tests, so we shouldn't have any problem supporting those.  At some point we might need a newer version too.

icraggs:
@waltronix 

> Suggested is an update of the required cmake version to 3.5 to support builds on ubuntu 16.04 LTS. This would also be beneficial for usage of OpenSSL on windows.

We don't need to update the required version of CMake to allow building on 16.04 LTS do we?  I'm building on 18.04 with CMake 3.10.2 just fine.

waltronix:
@icraggs  No, we do not need in update to be able to build on Ubuntu. But the more recent versions of cmake add additional features that might be helpful.
But I can try to keep the minimum version low until we encounter restrictions. Then we can decide by case.

icraggs:
So, is this done now?  Or is there more?

waltronix:
I'd like to pick this topic up again. But I didn't want to get into conflicts with the shared library topic.

icraggs:
I've finished with the CMake changes for static and shared library building in the develop branch now.

icraggs:
I've also added the definitions for calling the init functions for the shared libraries on Linux/MacOS.  And the testing of static builds.

[...]

PR #812, opened 2020-02-09
Use find package for openssl

waltronix:
With the targets provided by find_package, we can simply add the required libraries as PUBLIC and depending targets will simply know everything they need.

I changed OPENSSL_SEARCH_PATH to OPENSSL_ROOT_DIR, because this is the default from FindOpenSSL

This is meant as part of #799

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,13 @@
 sudo: required
 language: c
-compiler:
-- gcc
-- clang
-os:
-- linux
-- osx
-matrix:
-  exclude:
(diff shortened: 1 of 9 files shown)


waltronix:
Hi @icraggs,

would be cool, if you find time to have a look on the pull request.

Thanks
Andreas

icraggs:
Hi @waltronix.  Thanks, yes, I've been thinking about it - wondering if there's any effect that I haven't thought about yet.

PR #818, opened 2020-02-27
Fix serverURIs SSL connection handling

miks:
Signed-off-by: Miks Mikelsons miks.mikelsons@gmail.com


icraggs:
Thanks.  This isn't the complete solution though - that memory now isn't freed - there's a memory leak as shown in the test log.

icraggs:
I've added what I hope is a complete solution to the develop branch today.  Thanks for the PR.

miks:
Thanks, I can confirm that at least for me solution is working as expected

Issue #817, opened 2020-02-26
MEMLEAK: No way to work on HIGH_PERFORMANCE mode


AdrianMoranMontes:
**Describe the bug**
Running a simple main program (with v1.3.1) that creates client, connects it, publishes a message and disconnect, I saw that ASan sanitizer complies with messages like:

`SocketBuffer.c:82:20: runtime error: member access within misaligned address 0x606000000084 for type 'struct socket_queue', which requires 8 byte alignment`

Navigating through the code I see that malloc/realloc/free are redefined in `Heap.h` file, causing this misalignment. But I also see that there is a macro called HIGH_PERFORMANCE to disable heap tracking. Adding such define to the compiler, the misalignment issue is resolved (as expected) but a memory leak appears with asynchronous client. Each time a message is published with QoS 1 or 2, N bytes are leaked, being N the length of the used topic name.

**To Reproduce**
- Run any simple test with an Address Sanitizer to see misalignment complaints.
- Add HIGH_PERFORMANCE macro (`-DHIGH_PERFORMANCE=1`).
- Run simple test (that uses QoS 1 or QoS 2) again with a memory leak detector to see the leaks.

**Expected behavior**
- Heap tracking should not generate misalignment issues.
- Heap tracking is expected to be completely disabled when HIGH_PERFORMANCE macro is used.

**Environment**
 - OS: Ubuntu x64

**Other help**
Different parts of the code in `MQTTAsync.c` specifies that QoS 1 and QoS 2 releases the leaked bytes in the protocol. Going to `MQTTProtocolClient.c` file, in function `MQTTProtocol_storePublication`, such bytes should be released, but `Heap_findItem` function is used to check if the memory is in the heap (this is the only place where `Heap_findItem` function is used in Paho). Obviously, the return of this function is negative, since pure malloc has been used to allocate the memory, so the memory is never freed.

My first thought was to evaluate the QoS level instead of calling `Heap_findItem` function, but the `header` field of `publish` attribute is not set at this point of execution (which I suppose should be).

My second thought was to remove such checking of calling `Heap_findItem` (which works for my simple test program), but I have some doubts if I can always avoid such checking (I see that QoS 0 can also be stored in case of socket error).

**Additional**
`Tree.c` directly uses `mymalloc`/`myfree `alternatives for `malloc`/`free`, so if heap tracking disable is going to be supported, Tree.c file should also be changed.

icraggs:
[...]

PR #820, opened 2020-03-05
Fix for issue #817

AdrianMoranMontes:
See #817.

Fix for HIGH_PERFORMANCE macro still missing.


icraggs (changes requested):
The CI tests are failing, so this isn't working.

Also, where there is sizeof(double), it would be better to have sizeof(eyecatcher), assuming that works, which would be more future proof.  I realize I hadn't written it that way before, but I hadn't anticipated having to change the eyecatcher in the future either!

AdrianMoranMontes:
Hi Ian,
As CI reported out, there was issues in my fix that I think they are now corrected (tests passed locally).

Regarding the possibility to change the type of `eyecatcher`, I have defined a type for it, in a way that changing such `typedef` should be enough.

I hope CI tests pass this time.

icraggs (approved):
Looks good.  Thanks!

PR #823, opened 2020-03-10
Fix test 5 2a.

AdrianMoranMontes:
Minimal fix. Return code was not correctly evaluated for test 5 2a.

Regards.


icraggs:
Thanks.

PR #501, opened 2018-07-09
Full porting to mbedTLS library (2.13.0 version).

AdrianMoranMontes:
These changes allow to use mbedTLS library (2.13.0 version) as TLS layer instead of OpenSSL. To do this, internal SSLSocket API has been changed in order to work with a general struct (SSL handler) instead of with OpenSSL specific objects.

CMakeList has been modified to support two compilation alternatives:
`PAHO_WITH_OPENSSL to compile it with Openssl`
`PAHO_WITH_MBEDTLS to compile it with mbedTLS`

Openssl has been set (in CMake and in the code) as default TLS layer if both were available.

**This porting has been successfully tested on Linux and Windows.**

TODO:
- [ ] Select ciphers to be used in mbedTLS.
The field `enabledCipherSuites` of struct `MQTTClient_SSLOptions` is too tied to Openssl library. It should be necessary to translate the list of ciphers from Openssl syntax to mbedTLS or, better, define a Paho specific ciphersuites and traslate them to each library syntax.
Currently, default ciphers are used when mbedTLS is selected.

Signed-off-by: Adrian Moran <amoran@ikerlan.es>

PR #826, opened 2020-03-18
Porting to mbedTLS 2.16.5.

### LINKED ISSUE #1184: medTLS support
Is there any plans for supporting mbedTLS?

AdrianMoranMontes:
Related with old #501 (can discard and apply this PR).

This is a porting of Paho C library to mbedTLS library, tested with last version (2.16.5). All CI tests pass in my Linux machine, except the two ones related to de HA (I see @icraggs is currently working on them).

I would be pleasure to help on developing this new SSL layer for Paho. I know this PR is a first approach, but it could be a good base to start with.

Please, let me know if this work is of interest for community  in order to keep it updated.

Thank you.


icraggs:
I think it could definitely be of interest.  Do you use mbedTLS?  Are there advantages over OpenSSL?  

Does this PR enable mbedTLS to be used - so can we run the CI tests with these changes?

jumoog:
OpenSSL license is a nightmare for commerical use. 
mbedTLS is dual-licensed with the Apache License version 2.0  and GPLv2.

AdrianMoranMontes:
I know this version of Paho is not designed for embedded devices (we have https://github.com/eclipse/paho.mqtt.embedded-c for such purpose), but there are some cases in which the embedded device is not "so embedded" and it is better to use this "normal" Paho C library. However, OpenSSL is usually too big for such devices being mbedTLS much smaller than OpenSSL.

I think this porting will enable the use of Paho C is such scenarios.

Moreover, there are many developments that have mbedTLS in their framework and not OpenSSL (mbedTLS is much more portable than OpenSSL). Indeed, I performed this porting of Paho because of this need.

I would have liked to permorf this porting in a cleaner way (without so many "ifdef MBEDTLS"), but I didn't want to break current OpenSSL part. I think that the cleanest way would be to define a common SSL API (more or less what SSLSocket.h does) in a way that OpenSSL implementation could be performed in one .c file and mbedTLS in a separate .c file, letting CMake use one or another. However, I don't have enough knowledge about how Paho library internally works to do that by myself.

I have clear how mbedTLS handshake is performed, so if any of you wants to do this jobs, I would be pleasure of helping.


Regarding @jumoog question, you can download mbedTLS library (https://tls.mbed.org/download), compile it and then, compile Paho with my patch (replace `PAHO_WITH_SSL=TRUE` by `PAHO_WITH_MBEDTLS=TRUE` in cmake invokation). The next spets are the same. Indeed, I have been testing it with a copy of `travis-build.sh` script replacing that cmake defines. The tests run in the same way they run with OpenSSL. If you have issues to run them, let me know and I can explain in detail or even upload my modified script.

Thank you so much.

Adrian

Note: Provide `MBEDTLS_ROOT_DIR` in case mbedTLS is not located in common places, cmake needs to find it.

icraggs:
Is this applicable to Linux only?   Will the CI tests pass using mbedTLS?

I started merging to the latest develop branch and got as far as the static build building, but the CI tests not. The shared build failing at linking a sample.  This is with mbedTLS enabled.  On Ubuntu 18.04 with the Ubuntu mbedTLS package installed.

AdrianMoranMontes:
Hi @icraggs,
when I was working with this it worked both, in Linux/Windows and I would say that also with dynamic linkage (I am not 100% sure of this because the projects where I usually use Paho only support static linkage).

Now I am little bit busy, but I would like to help, so if you give me some steps to reproduce your errors I will give a try in the following days.

Thank you Ian.

PR #827, opened 2020-03-20
Allow proxy credentials with @ and special characters by decoding url-encoded http[s]_proxy

frbouvi:
Strong passwords often require special characters to be used. As the http[s]_proxy environment variable should be in the form http[s]://user:password@host:port/ , username or password containing /, :, @ can be misinterpreted and the proxy string incorrectly parsed. To avoid that, we allow to url encode proxy credentials. Eg if password is xxx@/%xxx, encode it like this : xxx%40%2F%25xxx

### DIFF
--- a/src/MQTTProtocolOut.c
+++ b/src/MQTTProtocolOut.c
@@ -140,8 +140,27 @@ int MQTTProtocol_connect(const char* ip_address, Clients* aClient, int websocket
 			basic_auth = (b64_data_t *)malloc(sizeof(char)*basic_auth_in_len);
 			basic_auth_in_len--;
 			p0 = (char *)basic_auth;
-			while(*p1 != '@')
-				*p0++ = *p1++;
+			/**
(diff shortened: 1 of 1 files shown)


icraggs:
I refactored the two cases into one function.  If you could check it still does as you need that would be very helpful.

frbouvi:
Hi, the refactor looks good to me. Thanks

De : Ian Craggs <notifications@github.com>
Envoyé : lundi 30 mars 2020 20:47
À : eclipse/paho.mqtt.c <paho.mqtt.c@noreply.github.com>
Cc : Frederic BOUVIER <frederic.bouvier@technilog.com>; Author <author@noreply.github.com>
Objet : Re: [eclipse/paho.mqtt.c] Allow proxy credentials with @ and special characters by decoding url-encoded http[s]_proxy (#827)


I refactored the two cases into one function. If you could check it still does as you need that would be very helpful.

PR #830, opened 2020-03-23
Fix high performance mode on Windows.

AdrianMoranMontes:
Issue on windows when HIGH_PERFORMANCE macro is enabled.

Regards,
Adrian


icraggs:
The issue being the creation of resources which are not used?  Rather than the execution failing?

AdrianMoranMontes:
Probably. As those mutexes are declared as extern I prefer to exclude  the senteces with ifdefs, but if they are not used, they can be removed.

Regards,
Adrian

Issue #857, opened 2020-04-23
MEMLEAK with HIGH_PERFORMANCE and memory increasing problem


AdrianMoranMontes:
**Describe the bug**
This issue is related with the closed #817 but it is slightly different.

In #817 I reported a memory leak when using HIGH_PERFORMANCE macro, which was kindly solved by @icraggs. However, I did not test the reception part of such fix until now (my fault).

If I enable HIGH_PERFORMANCE and receive a message on my client, I see three blocks of memory leaks. One, corresponding to the size of the topic name, other corresponding to the size of the payload and one last block corresponding to the malloc does in `Protocol_processPublication` of `MQTTAsync.c` to allocate memory for variable `mm`. If I follow the lifetime of `mm `variable, it is not freed anywhere!

If I disable HIGH_PERFORMANCE macro, the leaks disappear, but I see a biggest issue (hardest to detect). I tried to see where `mm` memory was freed without HIGH_PERFORMANCE but I could not find the place, what I saw, is that the memory consumed by mi client (process) increments each time it receives a MQTT message. The memory checker does not detect it as a leak, but the behavior is quite strange (I see a behavior like if we had something like a 'garbage collector' in the library ¿do we? :) ).

**To Reproduce**
Make simple program that subscribes an asynchronous client to a MQTT broker, monitor the system memory with `watch -n 1 free -k` and start publish messages from other client (`mosquitto_pub` in my case). It can be see how the system memory is decreasing each time a message arrives to the client. Other way is with the command `top`, and see how the memory consumed from our test program is increased each time a message arrives but never goes down.

I suggest to use big messages to see the effect easily. I used a file of ~25MB as payload.

All this process has been done in Linux, without SSL.

**Expected behavior**
Memory used to handle a received message should be freed after callback return (if user returned 1). Otherwise, the system will crash due to lack of memory.

AdrianMoranMontes:
My bad, I forgot MQTTAsync_freeMessage() call.

PR #860, opened 2020-04-24
Fix issue 857.

AdrianMoranMontes:
This fix solves issue #857.  Memory leaks are removes when HIGH_PERFORMANCE is used and memory does not increases with each received message until program crash when HIGH_PERFORMANCE is not used.

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -2961,8 +2961,12 @@ void Protocol_processPublication(Publish* publish, Clients* client, int allocate
 			MQTTPersistence_persistQueueEntry(client, (MQTTPersistence_qEntry*)qe);
 #endif
 	}
+	else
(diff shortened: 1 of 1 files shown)


icraggs (commented):
Hmm.  That storage is to be freed by the MQTTAsync_freeMessage() call, as in the paho_c_sub.c sample.

AdrianMoranMontes:
My bad @icraggs, I lost two day with this :(

I forgot the calls you mentioned. Sorry.

icraggs:
@AdrianMoranMontes Oh no.  I'm sorry to hear that.  

A long time ago the memory was freed automatically after the callback, but sometimes people wanted to process the message outside the callback, and this seemed the best solution.

Thank you for your contributions.

PR #861, opened 2020-04-24
src/CMakeLists.txt: fix build on uclibc and musl

ffontaine:
Build of version 1.3.2 with uclibc or musl fails on:

```
[ 58%] Linking C shared library libpaho-mqtt3c.so
/home/buildroot/autobuild/instance-1/output-1/host/opt/ext-toolchain/bin/../lib/gcc/arm-buildroot-linux-uclibcgnueabi/5.5.0/../../../../arm-buildroot-linux-uclibcgnueabi/bin/ld: cannot find -lanl
collect2: error: ld returned 1 exit status
```

This is due to commit 21ffebcb749fc8bd8578df85550f03f74e571620 which added a link to anl which is not available on uclibc or musl.

Fix this by checking if anl is found on the system before using it

Fixes:
 - http://autobuild.buildroot.org/results/4beb96d43180813906578b42875a1c3d4a905ed7

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


icraggs (changes requested):
Thanks.  It would help me remember why this has been done if a comment is added to the file to explain.

ffontaine:
I updated the PR as requested

icraggs (approved):
Thanks!

ffontaine:
Thanks for merging this PR however it seems that you have destroyed and recreated git release v1.3.2 to include this fix (among other ones). Doing so is generally seen as bad practice as users or other projects (such as buildroot) doesn't know anymore what is the "real" v1.3.2. It would have been better to release a v1.3.2.1 or wait for v1.3.3.

icraggs:
Ok, well I've done this before, generally for small build issues and no-one has mentioned anything.  There's always something that can be missed.  I'll try to avoid it in future.

dongqiceo:
what·s this ,lanl?

dongqiceo:
sorry,I cant find about anl lib info

ffontaine:
anl = asynchronous name lookup (https://github.com/bminor/glibc/blob/9fd3409842b3e2d31cff5dbd6f96066c430f0aa2/resolv/Makefile#L244)

PR #890, opened 2020-05-15
Fixed library names errors while compiling static samples

adwaitnd:
The library names used in the sample directory were paho-mqtt3as-static and paho-mqtt3cs-static. The actual names of the generated libraries are paho-mqtt3a-static and paho-mqtt3c-static. This pull changes the respective CMakeLists.txt files to make the samples compile successfully.

Tested on Ubuntu 18.04 amd64 with and without shared libraries being generated


icraggs (changes requested):
I think the paho_c_pub_static and other samples need the 's' versions of the libraries because those contain the SSL support.

icraggs:
As per my previous comment, I don't think this is correct.

PR #897, opened 2020-05-22
MQTTAsync_addCommand memory leak when using many clients and connecti…

### LINKED ISSUE #893: MQTTAsync_addCommand memory leak when using many clients and broker connection is not possible (i.e. network down)
**Describe the bug**
I am using version 1.3.2 of this library, slightly adapted to run in a hardware plattform not originally supported. Basically network, file handling, threading and critical section functions.

I compiled Paho library as static library and linked it to my application library.
I'm using MQTTAsync API since I want it to be thread-safe, as I intend that my application library handles multiple clients.

I tested with few clients, 2, 3 or even 4 and everything went fine. 
Then I started testing with a larger number, 50 clients running simoultaneously, connecting to the same broker (broker.hivemq.com:1883) but using different client ids and as long as I have an Internet  connection, the application works fine.

The issue arises when I start the clients and give them the connect command (only one call to MQTTAsync_connect per client) without having an Internet connection.
Memory consumption starts increasing until it eats all the available heap for my application library.

I've been doing some investigation and found out the problem comes from MQTTAsync_addCommand function (that is called at the end of MQTTAsync_connect)
In particular, the problem is that it keeps adding connect commands to the List.
The command passed by MQTTAsync_connect many times doesn't match with the first element in the list (because the first element corresponds to a different client) and as a consequence it gets inserted into the list, making it grow indefinitely.

**To Reproduce**
I'm not sure how you can reproduce this as I'm not using Windows or Linux OS, but one could try starting a large number of client threads, diconnecting laptop from network and check the size of the commands list.

**Expected behavior**
Commands should not be appended to the list if they are already in there.

**Fix**
Based on what I observed, I did a fix that seems to behave perfectly on my use case. If I try connecting all 50 client threads without network connection, consumed memory is still once they have finished allocating resources.
The fix basically consists on check if the command that we want to insert is already in the list. If it is, then we will remove it and free the associated memory and insert it again so that it goes to the top of the list.
This is the code:
```
/**
 * List callback function for comparing client handles and command types being CONNECT or DISCONNECT 
 * @param a first MQTTAsync_queuedCommand pointer
 * @param b second MQTTAsync_queuedCommand pointer
 * @return boolean indicating whether a and b are equal
 */
int compareClientConnectCommand(void* a, void* b)
{
	MQTTAsync_queuedCommand* cmd1 = (MQTTAsync_queuedCommand*)a;
	MQTTAsync_queuedCommand* cmd2 = (MQTTAsync_queuedCommand*)b;
	if (cmd1->client == cmd2->client)
	{
		if (cmd1->command.type == cmd2->command.type)
		{
			if (cmd1->command.type == CONNECT || cmd1->command.type == DISCONNECT)
			{
				return 1; //Item found in the list
			}
		}
	}
	return 0;	//Item NOT found in the list
}
static int MQTTAsync_addCommand(MQTTAsync_queuedCommand* command, int command_size)
{
	int rc = MQTTASYNC_SUCCESS;
	FUNC_ENTRY;
	MQTTAsync_lock_mutex(mqttcommand_mutex);
	/* Don't set start time if the connect command is already in process #218 */
	if ((command->command.type != CONNECT) || (command->client->c->connect_state == NOT_IN_PROGRESS))
		command->command.start_time = MQTTTime_start_clock();
	if (command->command.type == CONNECT ||
		(command->command.type == DISCONNECT && command->command.details.dis.internal))
	{
		MQTTAsync_queuedCommand* head = NULL;

		if (commands->first)
			head = (MQTTAsync_queuedCommand*)(commands->first->content);

		if (head != NULL && head->client == command->client && head->command.type == command->command.type)
			MQTTAsync_freeCommand(command); /* ignore duplicate connect or disconnect command */
		else
		{ 
			ListRemoveItem(commands, command, compareClientConnectCommand); /* remove command from the list if already there */
			ListInsert(commands, command, command_size, commands->first); /* add to the head of the list */
		}
	}
	else
	{
		ListAppend(commands, command, command_size);
#if !defined(NO_PERSISTENCE)
		if (command->client->c->persistence)
			MQTTAsync_persistCommand(command);
#endif
	}
	MQTTAsync_unlock_mutex(mqttcommand_mutex);
#if !defined(_WIN32) && !defined(_WIN64)
	rc = Thread_signal_cond(send_cond);
	if (rc != 0)
		Log(LOG_ERROR, 0, "Error %d from signal cond", rc);
#else
	rc = Thread_post_sem(send_sem);
#endif
	FUNC_EXIT_RC(rc);
	return rc;
}
```
I hope you can take a look and see if this makes sense for you or it should not happen in the first place.

Best regards,
Alejandro

alex-gimenez:
…on to broker is not possible. Issue #893

Signed-off-by: Alejandro Giménez <alexx091@gmail.com>

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -1168,7 +1168,29 @@ static int MQTTAsync_restoreCommands(MQTTAsyncs* client)
 }
 #endif
 
-
+/**
+ * List callback function for comparing client handles and command types being CONNECT or DISCONNECT 
+ * @param a first MQTTAsync_queuedCommand pointer
+ * @param b second MQTTAsync_queuedCommand pointer
+ * @return boolean indicating whether a and b are equal
+ */
+static int clientCompareConnectCommand(void* a, void* b)
+{
+	MQTTAsync_queuedCommand* cmd1 = (MQTTAsync_queuedCommand*)a;
+	MQTTAsync_queuedCommand* cmd2 = (MQTTAsync_queuedCommand*)b;
+	if (cmd1->client == cmd2->client)
+	{
+		if (cmd1->command.type == cmd2->command.type)
+		{
+			if (cmd1->command.type == CONNECT || cmd1->command.type == DISCONNECT)
+			{
+				return 1; //Item found in the list
+			}
+		}
+	}
+	return 0;	//Item NOT found in the list
+}								   
+ 
 static int MQTTAsync_addCommand(MQTTAsync_queuedCommand* command, int command_size)
 {
 	int rc = MQTTASYNC_SUCCESS;
@@ -1189,7 +1211,10 @@ static int MQTTAsync_addCommand(MQTTAsync_queuedCommand* command, int command_si
 		if (head != NULL && head->client == command->client && head->command.type == command->command.type)
 			MQTTAsync_freeCommand(command); /* ignore duplicate connect or disconnect command */
 		else
+		{ 
+			ListRemoveItem(commands, command, clientCompareConnectCommand); /* remove command from the list if already there */
 			ListInsert(commands, command, command_size, commands->first); /* add to the head of the list */
+		}
 	}
 	else
 	{


icraggs:
This has now been merged to the develop branch.  Thanks.

PR #912, opened 2020-06-19
MQTT on WS fixes

KrzysztofC-RST:
Because of problems detected during using Paho when MQTT server was sending massive amount of data by WS (Web Socket) fames, I made some research for reasons for that.
There are some description on provided new commits, but generally, I may say, that problem was caused by cutting frames and packing few MQTT frames into WS frame by sending serwer. This is correct from the server side and it may happen on specific server settings. Client side (Paho) in some cases not collecting correctly received data as MQTT frames. This fixes solve this problems.

### DIFF
--- a/src/MQTTPacket.c
+++ b/src/MQTTPacket.c
@@ -112,6 +112,8 @@ void* MQTTPacket_Factory(int MQTTVersion, networkHandles* net, int* error)
 	FUNC_ENTRY;
 	*error = SOCKET_ERROR;  /* indicate whether an error occurred, or not */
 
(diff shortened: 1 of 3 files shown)


icraggs:
Thanks.  Can you tell me what server was doing this?

KrzysztofC-RST:
Hello Ian. It is EMQX v4.1.0 https://github.com/emqx/emqx
It's configured for big performance in some way. Older EQMX version, used in our project, was not generating this problems, but there was no such a performance optimalizations.

icraggs:
Ok, thanks.  I don't think I've yet seen any other broker do that, although it was always possible of course.

Issue #745, opened 2019-10-11
Stack smashing


fpagliughi:
I've been having a recurring issue with a crash of my sample apps in Linux (Ubuntu 18.04 at the moment) where it complains about stack smashing. 

```
*** stack smashing detected ***: <unknown> terminated
Aborted (core dumped)
```

I believe that it happens when I switch between different versions of the Paho C library during development, so I don't think it's a problem most users would commonly see.

I believe that it has something to do with the default file persistence used by the library. One guess may be that if the library tries to access the persistence file created from a different version of the lib, then maybe it gets some bad data and things go ugly. If that's the case, then perhaps the file persistence could use a version tag written to the file?

Here's a max dump when trying to publish a message:

```
Publishing a message on the 'test' topic
20191011 145742.921 (2722629376)  (1)> MQTTAsync_cycle:3742
20191011 145742.921 (2722629376)   (2)> Socket_getReadySocket:237
20191011 145742.921 (2731022080)  (1)< MQTTAsync_checkTimeouts:1852
20191011 145742.921 (2731022080)  (1)> Thread_wait_cond:413
20191011 145742.921 (2753394816) (0)> MQTTAsync_sendMessage:3546
20191011 145742.921 (2753394816)  (1)> MQTTAsync_send:3477
20191011 145742.921 (2753394816)   (2)> UTF8_validateString:157
20191011 145742.921 (2753394816)    (3)> UTF8_validate:131
20191011 145742.921 (2753394816)    (3)< UTF8_validate:143 (1)
20191011 145742.921 (2753394816)   (2)< UTF8_validateString:159 (1)
20191011 145742.921 Allocating 160 bytes in heap at file /home/fmp/mqtt/paho-rust/paho-mqtt-sys/paho.mqtt.c/src/MQTTAsync.c line 3509 ptr 0x564484006b30

20191011 145742.921 Allocating 16 bytes in heap at file /home/fmp/mqtt/paho-rust/paho-mqtt-sys/paho.mqtt.c/src/MQTTProtocolClient.c line 874 ptr 0x564484006eb0

20191011 145742.921 (2753394816)   (2)> MQTTStrncpy:851
20191011 145742.921 (2753394816)   (2)< MQTTStrncpy:861
[...]

PR #921, opened 2020-07-19
Added async continuous publisher example

fpagliughi:
On a user request a while back, I created examples in the Paho C++ and Rust libraries to demonstrate how to (somewhat) periodically sample data and publish continuously as would be done in a typical MQTT app that reads from a sensor. I created a sample that reads the system clock in millisecond or greater resolution and publishes that to the MQTT broker as simulated data readings, and tested with rates of 100 - 1000 msg/sec.

It was those samples that originally showed the stack-smashing bug (#745) when the app was allowed to run on a PC for five or ten hours. This example was a re-write of those sample apps in C to check if the bug was in the C lib or wrapper libs.

It's presented here if you want to include it as a sample or just test. So far, compiled against the `develop` branch and running for 24hrs, there is no stack-smash. Very nice.


fpagliughi:
Oops, the Windows build failed. I'll try to find a Windows machine and get it fixed.

Issue #927, opened 2020-07-29
Client doesn't send PINGREQ


jumoog:
- OS: Windows Server 2012 R2
 - Version Pahoo 1.3.4

The Client should send PINGREQ but it never sends it. The function MQTTProtocol_keepalive is never called. Any Ideas whats wrong? eclipse mqtt-spy works fine

[log.txt](https://github.com/eclipse/paho.mqtt.c/files/4994514/log.txt)

icraggs:
I just tried Windows 10 and it works fine for me (the develop branch).  I don't have any older versions of Windows to hand except those that are used in the CI tests.  

If MQTTProtocol_keepalive is not being called, it depends on the statement:

    if (MQTTTime_difftime(now, last) > (retryLoopInterval * 1000))

so probably the behaviour of MQTTTime_difftime().  That depends on the GetTickCount function -- hard to see what could go wrong there.

jumoog:
it looks like MQTTAsync_retry gets called as well

jumoog:
```
20200729 173456.971 MQTTAsync_retry
20200729 173457.663 now <3656204734>
20200729 173457.663 last <0>
20200729 173457.663 diff <3656204734>
20200729 173457.663 retryLoopInterval <5000>
```
for some reason the compare is not true and else always gets called

GetTickCount overflows every every 49 days. MS recommends to use GetTickCount64

jumoog:
I switched to GetTickCount64 and now everything works

icraggs:
I'd assumed it was just going to be a problem at the instant of rollover.

jumoog:
yes 42 days

![grafik](https://user-images.githubusercontent.com/5268791/88828812-05428300-d1cc-11ea-80fe-9df55fcfffee.png)

icraggs:
Does GetTickCount fail completely now then?  I had thought from the description that it might rollover and start from 0 again.  Or something like that.

jumoog:
@icraggs  it was like this and went always into else branch
`if (MQTTTime_difftime(3656204734, 0) > (5* 1000))`

jumoog:
Maybe the Type for `MQTTTime_difftime` and `MQTTTime_elapsed` is wrong? long (range: –2147483648 through 2147483647 decimal) but `GetTickCount()` returns a DWORD (range: 0 through 4294967295 decimal)

jumoog:
[...]

PR #928, opened 2020-07-29
use GetTickCount64 instead of GetTickCount

jumoog:
GetTickCount overflows roughly every 49 days.
Code that does not take that into account can loop indefinitely.  
GetTickCount64 operates on 64 bit values and does not have that problem.

fix for #927

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -92,7 +92,7 @@ void MQTTAsync_global_init(MQTTAsync_init_options* inits)
 void MQTTAsync_init_rand(void)
 {
 	START_TIME_TYPE now = MQTTTime_start_clock();
-	srand(now);
+	srand((UINT)now);
 }
 #elif defined(AIX)
 void MQTTAsync_init_rand(void)
@@ -4320,7 +4320,11 @@ int MQTTAsync_waitForCompletion(MQTTAsync handle, MQTTAsync_token dt, unsigned l
 {
 	int rc = MQTTASYNC_FAILURE;
 	START_TIME_TYPE start = MQTTTime_start_clock();
+#if defined(_WIN32) || defined(_WIN64)
+	ULONGLONG elapsed = 0L;
+#else
 	unsigned long elapsed = 0L;
+#endif
 	MQTTAsyncs* m = handle;
 
 	FUNC_ENTRY;
--- a/src/MQTTClient.c
+++ b/src/MQTTClient.c
@@ -351,15 +351,15 @@ static int MQTTClient_cleanSession(Clients* client);
 static MQTTResponse MQTTClient_connectURIVersion(
 	MQTTClient handle, MQTTClient_connectOptions* options,
 	const char* serverURI, int MQTTVersion,
-	START_TIME_TYPE start, long millisecsTimeout,
+	START_TIME_TYPE start, TIME_OUT_TYPE millisecsTimeout,
 	MQTTProperties* connectProperties, MQTTProperties* willProperties);
 static MQTTResponse MQTTClient_connectURI(MQTTClient handle, MQTTClient_connectOptions* options, const char* serverURI,
 	MQTTProperties* connectProperties, MQTTProperties* willProperties);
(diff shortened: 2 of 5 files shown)


icraggs (commented):
Thanks.  GetTickCount64 is first available on Windows Vista, so we'll have to use GetTickCount on earlier OSes:

#if WINVER >=_WIN32_WINNT_VISTA
	return GetTickCount64();
#else
	return GetTickCount();
#endif

jumoog:
Windows 7 is EOL. But seriously I don't know how to solve the problem on older operating systems

fpagliughi:
Windows Vista came out 14 years ago and went EOL three or four years ago. If company hasn't upgraded their OS in that time, are they really doing a lot of new development on them? And even if so, should we kill ourselves to support them when Microsoft won't even do it anymore?

I would vote for dropping support for anything prior to Vista at this point. That OS actually introduced a number of good low-level features that had been lacking in the Microsoft API. Condition variables comes to mind. It could be useful to eventually use some of these things to clean up the Microsoft code in here.

But, as usual, I would normally do this at a good point in the release cycle. Have a solid version that can still be used by older systems, then make the leap. For example, merge this in when we go to Release 1.4.0. Then if someone needs support for XP or older, we can point them to "any release" before that.

icraggs:
I know people are using it on XP - so I'd rather not put any more obstacles in their way.  No one else has reported the issue, so I'm content with defaulting to GetTickCount for OSes earlier than Vista.  There are potential solutions for that, but this way narrows down the set of failing cases anyway.

I'd started writing the fix this (dual) way when Kilian submitted this PR.  As far as I'm concerned, it's the path of least maintenance work in the future.  I'm happy to do it, if no one else wants to!

jumoog:
wait i forgot something! MQTTClient.c
//EDIT:
MQTTClient.c is hard!
mixing timeout (long) and times (ULONGLONG)

jumoog:
Severity	Code	Description	Project	File	Line	Suppression State
Warning	C4244	'=' : conversion from 'ULONGLONG' to 'long', possible loss of data	paho-mqtt3c	C:\Repo\paho.mqtt.c\src\MQTTClient.c	2467
`tp.tv_sec = timeout / 1000;`
Severity	Code	Description	Project	File	Line	Suppression State
Warning	C4244	'function' : conversion from 'ULONGLONG' to 'int', possible loss of data	paho-mqtt3c	C:\Repo\paho.mqtt.c\src\MQTTClient.c	2572
`if ((*rc = Thread_wait_sem(m->connect_sem, timeout)) == 0)`

jumoog:
> 
> 
> I know people are using it on XP - so I'd rather not put any more obstacles in their way. No one else has reported the issue, so I'm content with defaulting to GetTickCount for OSes earlier than Vista. There are potential solutions for that, but this way narrows down the set of failing cases anyway.
> 
> I'd started writing the fix this (dual) way when Kilian submitted this PR. As far as I'm concerned, it's the path of least maintenance work in the future. I'm happy to do it, if no one else wants to!

it's okay, i'll add it once i solved the Problems with MQTTClient.c

src/MQTTAsync.c
```diff
@@ -92,7 +92,7 @@ void MQTTAsync_global_init(MQTTAsync_init_options* inits)
 void MQTTAsync_init_rand(void)
 {
 	START_TIME_TYPE now = MQTTTime_start_clock();
-	srand(now);
+	srand((UINT)now);
```
jumoog, line 95: good?

jumoog:
no build warnings anymore. If i add checks for pre Vista it has no effect! All build system are newer and ifdef only affects the compiler. Curl made this: https://github.com/curl/curl/commit/1fc0617dccc3fa138235f219e2eaa7b405d1162e

icraggs:
@jumoog yes all the AppVeyor build systems are newer.  There isn't an option for earlier, unsurprisingly.  So it relies on other means to test it.  Either by forcing the use of GetTickCount() by temporarily changing the #ifdefs or running it on a real old OS.  Which I don't have so that depends on other people.  The risk should be low, I think.

PR #930, opened 2020-07-31
use proper type for time functions

jumoog:
long has a different range

tested at system with 44 Days and 12h uptime with Tickcount: **3844828906**

I still perfer a version with no overflow.


mtrensch:
Looks like a little typo:
MQTT_TIME_TPYE should be MQTT_TIME_TYPE

PR #932, opened 2020-08-06
Fix debian package

klaussnd:
Currently making a Debian package is broken, setting `PAHO_BUILD_DEB_PACKAGE=ON` makes the build fail. There are currently two conflicting ways to build a Debian package which break each other. This PR fixes creating a package with `cpack` and removes the conflicting way. It also cleans up the `cpack` configuration by removing the unnecessary dependency on `cmake-debhelper` (the required init scripts are already generated by `cpack` by default) and fixing shlibdeps (which is only executed if the list with explicit dependencies is not set).


icraggs:
Thank you very much.

PR #939, opened 2020-08-27
Fix compilation when running make with multiple jobs (make -j)

aler9:
Hello, if i try to compile the library with `make -j$(nproc)`, the process fails:
```
/usr/bin/ld: cannot find -lpaho-mqtt3cs
collect2: error: ld returned 1 exit status
/usr/bin/ld: cannot find -lpaho-mqtt3cs
collect2: error: ld returned 1 exit status
```

This is caused by the fact that the samples are built before the library.
This patch solves the issue by setting the right dependencies of the samples in the Makefile, and allows to use `make` with the -j flag, reducing compilation times (when cross-compiling, compilation time changed from 4mins to 30secs).


icraggs:
Thank you.

PR #948, opened 2020-09-07
add missing alpn support to async

jumoog:
in 9e9cba2ea2f8a515a9102e94961dc6e4808bb7e3 only the synchronous client got the new feature.

It should work for async as well. But i found an Android bug: https://android.googlesource.com/platform/external/conscrypt/+/f8a9b546d57c4731805e73e1e96ff2fb3e77d6e0%5E!/ 

```
Calling SSL_CTX_set_alpn_protos appears to be detrimental to thread
safety since the implementation of it resets the values. It's not
idempotent to call it multiple times like SSL_CTX_enable_npn.
```


icraggs:
Thanks - I'd missed that.   I expect there to be other issues on Android as well as I've never tested this client there, so for now at least, I'm happy to include this.

PR #951, opened 2020-09-14
Adding PKCS#11 Support

cstaylor:
This will add optional support for the [PKCS#11](https://en.wikipedia.org/wiki/PKCS_11) interface with openssl.  Users can leverage this to work with any HSM that supports the PKCS#11 interface. 

Tests bring in [SoftHSM2](https://github.com/opendnssec/SoftHSMv2)
Signed-off-by: Christopher Taylor <cstaylor@argo.ai>


cstaylor:
As far as I can tell, the test that failed was a timeout on test 107 on the Mac?  Are there any additional logs that might explain it?  Can we just retry the mac tests?

robpottify:
This looks like great work @cstaylor . 
Are you still planning on merging?

test/pkcs11_hsm_test/libp11_readhsm.c
```diff
@@ -0,0 +1,291 @@
+/*
+ * Copyright © 2020, Andreas Jellinghaus <andreas@ionisiert.de>
+ * All rights reserved.
```
icraggs: This is a file which isn't written by you and under a different license?  If so, it will have to be declared separately or removed.

src/SSLSocket.c
```diff
@@ -541,6 +551,188 @@ static unsigned int call_ssl_psk_cb(SSL *ssl, const char *hint, char *identity,
...
+    slots = sslopts->pkcs11_slots;
+    nslots = sslopts->pkcs11_slot_num;
+    tokenlabel = sslopts->tokenLabel;
+    keylabel = sslopts->keyLabel;
+    pinvalue = sslopts->pinValue;
+
+	slot = PKCS11_find_token(pkcs11_ctx, slots, nslots);
+    for (unsigned int i = 1; i <= nslots; i++) {
+        if (slot == NULL || slot->token == NULL) {
+            fprintf(stderr, "no token!\n");
```
icraggs, line 576: Why is there output to stderr rather than the log?  Is this a utility/test or similar?

icraggs (changes requested):
Thanks for the comprehensive PR.  Incorporating it will take a while, I expect adding the tests to the CI suite will be a substantial part of that.

test/pkcs11_hsm_test/libp11_readhsm.c
cstaylor: Fixed!

src/SSLSocket.c
cstaylor: Yeah this was left over from some testing we had done (based on code from the p11 project).  I've removed the `printf`s

PR #979, opened 2020-10-20
Don't call the failure callback once connected

### LINKED ISSUE #972: Connect onSuccess/onFailure can be called more than once
I have a consistent crash in my program which started occurring when I switched from using `QoS=1` to `QoS=0`. I'm using the async client by means of the `paho.mqtt.rust` crate (using the master branch). While I understand that the Rust crate is a project on it's own, I traced the problem back to the underlying `paho.mqtt.c` package.

I tried it with the version bundled by the Rust crate (1.3.2), as well as with the latest version (1.3.6) build locally by myself. Both give the exact same traces.

My Rust is pretty decent, but my C is not. So it's a challenge for me to further debug this myself, hence I'm hoping this could be checked by someone with some more C skills.  

I have no clear way to reproduce the issue stand alone, but as you can see in the logs (trace log using the environment variable as described in the readme and a GDB backtrace) it actually happens right after the first publish:

[Trace log](https://gist.github.com/svanharmelen/da045215f350fbea6c74c2648b9092da)
[GBD log](https://gist.github.com/svanharmelen/3e1ce30e000e0a291d289f69787c1e68)

Please let me know if there is anything I can/should test that might help understand the problem better. Thanks!

svanharmelen:
If a connection failure occurs any time after a connection succeeded
and triggered the success callback, it should not try to also call the
failure callback as it would do this using the same context which was
already used when making the success callback.

Fixes eclipse#972

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -1395,7 +1395,7 @@ static int MQTTAsync_unpersistInflightMessages(Clients* c)
 #endif
 
 /**
- * List callback function for comparing client handles and command types being CONNECT or DISCONNECT 
+ * List callback function for comparing client handles and command types being CONNECT or DISCONNECT
  * @param a first MQTTAsync_queuedCommand pointer
  * @param b second MQTTAsync_queuedCommand pointer
  * @return boolean indicating whether a and b are equal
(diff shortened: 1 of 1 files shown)


svanharmelen:
A same fix is already added to the develop branch a few hours ago, so I'll close this PR again 👍🏻

PR #983, opened 2020-10-23
Disconnect never occurs if QoS 2 messages are still in progress

vm-vloz:
Attached a trace file that shows the problem:
Immediately before disconnecting, the application sends a message with QoS 2.
The application is still able to send messages after multiple disconnect calls, and the disconnect callbacks are never invoked.
The connect_state is not getting set to DISCONNECTING, so the MQTTAsync_checkTimeouts call does not result in a call to MQTTAsync_checkDisconnect.

[mqtt-disconnect-log.txt](https://github.com/eclipse/paho.mqtt.c/files/5431636/mqtt-disconnect-log.txt)

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -2120,7 +2120,6 @@ static int MQTTAsync_processCommand(void)
 		{
 			if (command->client->c->connect_state != NOT_IN_PROGRESS)
 			{
-				command->client->c->connect_state = DISCONNECTING;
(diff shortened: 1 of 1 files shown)


icraggs:
Thanks.  You've tried this out, presumably?

vm-vloz:
> Thanks. You've tried this out, presumably?

Yes. This fixes the problem I described.

icraggs:
Excellent.  Thanks.

PR #984, opened 2020-10-25
Works now without warnings under MinGW32 under Windows

CapitanFuture:
Second try for commit message with signed off


icraggs (changes requested):
The statement on line 62 should be #else, not #elif - causes the windows builds to fail.

icraggs:
Implemented in the develop branch.

PR #994, opened 2020-11-05
Memory leak corrections

deonmcclung:
Here are some fixes for some memory leaks that I found while using the Async library.

**A couple of them were active leaks of the topic name.** Others were places where an owning pointer goes out of scope without freeing or transferring ownership to another pointer. These are potential leaks that I corrected, though I never saw them actively leaking.

### DIFF
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -860,7 +860,7 @@ static int MQTTAsync_unpersistInflightMessages(Clients* c)
 #endif
 
 /**
- * List callback function for comparing client handles and command types being CONNECT or DISCONNECT 
+ * List callback function for comparing client handles and command types being CONNECT or DISCONNECT
  * @param a first MQTTAsync_queuedCommand pointer
  * @param b second MQTTAsync_queuedCommand pointer
  * @return boolean indicating whether a and b are equal
@@ -880,7 +880,7 @@ static int clientCompareConnectCommand(void* a, void* b)
 		}
 	}
 	return 0;	//Item NOT found in the list
-}								   
+}
 
 
 int MQTTAsync_addCommand(MQTTAsync_queuedCommand* command, int command_size)
@@ -903,7 +903,7 @@ int MQTTAsync_addCommand(MQTTAsync_queuedCommand* command, int command_size)
 		if (head != NULL && head->client == command->client && head->command.type == command->command.type)
 			MQTTAsync_freeCommand(command); /* ignore duplicate connect or disconnect command */
 		else
-		{ 
+		{
 			ListRemoveItem(MQTTAsync_commands, command, clientCompareConnectCommand); /* remove command from the list if already there */
 			ListInsert(MQTTAsync_commands, command, command_size, MQTTAsync_commands->first); /* add to the head of the list */
 		}
@@ -2057,6 +2057,12 @@ static void MQTTAsync_freePack(MQTTPacket* pack)
 		else if (pack->header.bits.type == DISCONNECT)
 		{
 			MQTTPacket_freeAck((Ack*)pack);
+		}
+		else if (pack->header.bits.type == PINGRESP ||
+			     pack->header.bits.type == PINGREQ)
+		{
+			// No free(pack) here as pack comes from MQTTPacket_header_only, which returns a
+			// reference to a static variable.
 		}
 		else
 			free(pack);
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -3119,7 +3119,7 @@ static MQTTPacket* MQTTAsync_cycle(int* sock, unsigned long timeout, int* rc)
 				}
 			}
 			else if (pack->header.bits.type == PUBREL)
-				*rc = MQTTProtocol_handlePubrels(pack, *sock);
+				*rc = MQTTProtocol_handlePubrels(pack, *sock, 1);
 			else if (pack->header.bits.type == PINGRESP)
 				*rc = MQTTProtocol_handlePingresps(pack, *sock);
 			else
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -2037,6 +2037,32 @@ static int MQTTAsync_completeConnection(MQTTAsyncs* m, Connack* connack)
 	return rc;
 }
 
+static void MQTTAsync_freePack(MQTTPacket* pack)
+{
+	FUNC_ENTRY;
+	if (pack)
+	{
+		if (pack->header.bits.type == CONNACK)
+		{
+			MQTTPacket_freeConnack((Connack*)pack);
+		}
+		else if (pack->header.bits.type == SUBACK)
+		{
+			MQTTPacket_freeSuback((Suback*)pack);
+		}
+		else if (pack->header.bits.type == UNSUBACK)
+		{
+			MQTTPacket_freeUnsuback((Unsuback*)pack);
+		}
+		else if (pack->header.bits.type == DISCONNECT)
+		{
+			MQTTPacket_freeAck((Ack*)pack);
+		}
+		else
+			free(pack);
+	}
+	FUNC_EXIT;
+}
 
 /* This is the thread function that handles the calling of callback functions if set */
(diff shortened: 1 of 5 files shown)


src/MQTTAsyncUtils.c
```diff
@@ -2584,9 +2620,9 @@ void Protocol_processPublication(Publish* publish, Clients* client, int allocate
 		if (client->persistence)
 			MQTTPersistence_persistQueueEntry(client, (MQTTPersistence_qEntry*)qe);
 #endif
+		publish->topic = NULL; /* Prevent topic name free as reference was passed to qe */
 	}
 exit:
-	publish->topic = NULL;
```
deonmcclung, line 2586: Moving this line inside the condition above fixes an active leak of the topic name.

src/MQTTProtocolClient.c
```diff
@@ -583,7 +583,10 @@ int MQTTProtocol_handlePubrels(void* pack, int sock)
 			if (m->MQTTVersion >= MQTTVERSION_5)
 				MQTTProperties_free(&m->properties);
 			if (m->publish)
+			{
 				ListRemove(&(state.publications), m->publish);
+				free(m->publish->topic);
```
deonmcclung, line 588: This free fixes an active leak of the topic name.

src/MQTTAsyncUtils.c
```diff
@@ -3073,7 +3109,7 @@ static MQTTPacket* MQTTAsync_cycle(int* sock, unsigned long timeout, int* rc)
 								data.token = command->command.token;
 								data.reasonCode = ackrc;
 								data.properties = msgprops;
-								data.packet_type = pack->header.bits.type;
+								data.packet_type = packet_type;
```
deonmcclung, line 3115: Changing this to get the packet type from a local variable instead of from _pack->header.bits.type_ allows _pack_ to be freed at a point before this in a consistent manner. Other changes to use this variable above are for convenience/efficiency.

src/MQTTAsyncUtils.c
```diff
@@ -2037,6 +2037,32 @@ static int MQTTAsync_completeConnection(MQTTAsyncs* m, Connack* connack)
 	return rc;
 }
 
+static void MQTTAsync_freePack(MQTTPacket* pack)
```
deonmcclung, line 2040: Depending on type that pack represents, other data members may need to be freed. Call the appropriate functions to do so.

src/MQTTAsyncUtils.c
```diff
@@ -2058,23 +2084,31 @@ thread_return_type WINAPI MQTTAsync_receiveThread(void* n)
 		pack = MQTTAsync_cycle(&sock, timeout, &rc);
 		MQTTAsync_lock_mutex(mqttasync_mutex);
 		if (MQTTAsync_tostop)
+		{
+			MQTTAsync_freePack(pack);
```
deonmcclung, line 2094: _pack_ is going out of scope without transfer. (pack may or may not point to memory)

src/MQTTAsyncUtils.c
```diff
@@ -2058,23 +2084,31 @@ thread_return_type WINAPI MQTTAsync_receiveThread(void* n)
...
 			break;
+		}
 		timeout = 1000L;
 
 		if (sock == 0)
+		{
+			MQTTAsync_freePack(pack);
```
deonmcclung, line 2101: _pack_ is going out of scope without transfer. (pack may or may not point to memory)

src/MQTTAsyncUtils.c
```diff
@@ -2058,23 +2084,31 @@ thread_return_type WINAPI MQTTAsync_receiveThread(void* n)
...
 			continue;
+		}
 		/* find client corresponding to socket */
 		if (ListFindItem(MQTTAsync_handles, &sock, clientSockCompare) == NULL)
 		{
 			Log(TRACE_MINIMUM, -1, "Could not find client corresponding to socket %d", sock);
 			/* Socket_close(sock); - removing socket in this case is not necessary (Bug 442400) */
+			MQTTAsync_freePack(pack);
```
deonmcclung, line 2109: _pack_ is going out of scope without transfer. (pack may or may not point to memory)

src/MQTTAsyncUtils.c
```diff
@@ -2058,23 +2084,31 @@ thread_return_type WINAPI MQTTAsync_receiveThread(void* n)
...
 			continue;
 		}
 		m = (MQTTAsyncs*)(MQTTAsync_handles->current->content);
 		if (m == NULL)
 		{
 			Log(LOG_ERROR, -1, "Client structure was NULL for socket %d - removing socket", sock);
 			Socket_close(sock);
+			MQTTAsync_freePack(pack);
```
deonmcclung, line 2117: _pack_ is going out of scope without transfer. (pack may or may not point to memory)

src/MQTTProtocolOut.c
```diff
@@ -341,6 +341,7 @@ int MQTTProtocol_handlePingresps(void* pack, int sock)
 	client = (Clients*)(ListFindItem(bstate->clients, &sock, clientSocketCompare)->content);
 	Log(LOG_PROTOCOL, 21, NULL, sock, client->clientID);
 	client->ping_outstanding = 0;
+	free(pack);
```
deonmcclung, line 344: According to a comment on MQTTAsyncUtils.c:3028, handle functions are supposed to free the pack memory when finished. All the others do, this one did not.

src/MQTTAsyncUtils.c
```diff
@@ -3086,7 +3122,7 @@ static MQTTPacket* MQTTAsync_cycle(int* sock, unsigned long timeout, int* rc)
 				}
 			}
 			else if (pack->header.bits.type == PUBREL)
-				*rc = MQTTProtocol_handlePubrels(pack, *sock);
+				*rc = MQTTProtocol_handlePubrels(pack, *sock, 1);
```
deonmcclung, line 3128: To avoid a leak of the Topic Name, MQTTProtocol_handlePubrels must delete the topic name in addition to the pack upon its completion.

src/MQTTClient.c
```diff
@@ -2551,7 +2551,7 @@ static MQTTPacket* MQTTClient_cycle(int* sock, ELAPSED_TIME_TYPE timeout, int* r
 				*rc = MQTTProtocol_handlePubrecs(pack, *sock);
 			}
 			else if (pack->header.bits.type == PUBREL)
-				*rc = MQTTProtocol_handlePubrels(pack, *sock);
+				*rc = MQTTProtocol_handlePubrels(pack, *sock, 0);
```
deonmcclung, line 2563: The MQTTClient interface expects the caller of MQTTClient_receive to free the topicName and message, so this function is to not delete the topicName.

icraggs:
If I run these changes with the PAHO_HIGH_PERFORMANCE CMake option, some of the CI tests fail.  For instance: 

"paho.mqtt.c/build.paho/test/test5 "--test_no" "12" "--ws" "--hostname" "localhost" "--client_key" "paho.mqtt.c/test/ssl/client.pem" "--capath" "paho.mqtt.c/test/ssl/capath"

icraggs (changes requested):
A significant number of the CI tests fail (on Linux) if they are run with the CMake option PAHO_HIGH_PERFORMANCE, which disables the internal memory tracking and so detects some memory errors more directly.

deonmcclung:
Thanks. I'll look into those. It is a little odd because PAHO_HIGH_PERFORMANCE is how I generally run, but I'll check those out.

deonmcclung:
So, I'm having trouble running the tests even on the develop branch (or at least they are not passing). I am running with:

EmqX broker listening on port 1883.
Ubuntu on WSL

```
cmake -DPAHO_HIGH_PERFORMANCE=TRUE -DPAHO_WITH_SSL=TRUE ..
make
python test/mqttsas.py &
```
Is there anything else that I need to do to get them to run cleanly, even  before my changes?

icraggs:
Hmm.  I missed this.  I use the travis-build.sh shell script included with the addition of -DPAHO_HIGH_PERFORMANCE=TRUE  to the CMake command, after setting the environment variables:

     export PAHO_BUILD_SHARED=TRUE
     export PAHO_BUILD_STATIC=TRUE

PR #1003, opened 2020-11-26
Generate shlibs control file for debian package

elrafoon:
Without the control file, dpkg-shlibdeps won't recognize this package.
Depending packages created by library user would end up without
dependency on libpaho-mqtt.

Signed-off-by: Stanislav Ravas <ravas@tind.sk>


icraggs:

PR #1004, opened 2020-11-26
Replace symlinks with real files to fix pacman package build

elrafoon:
When packaging paho.mqtt.c with MinGW for MSYS2, pacman unable is extract the symlinks from source archive - it uses bsdtar and it fails on symlinks when running under windows.

So patching sources in the package is not an option, because sources don't even get unpacked.

Only solution I found is to replace the symlinks with real files.


icraggs (changes requested):
These links were created by the c_rehash utility, as you probably know.  If we are going to do this, then I think we should have a readme in the directory explaining how and why we've done this.  And how to recreate should the certificates need changing (which they will at some point).

elrafoon:
To be honest, I didn't know about c_rehash utility nor this project is using it. So I looked up the manpage.

In first sections it says:

> c_rehash scans directories and calculates a hash value of each .pem, .crt, .cer, or .crl file in the specified directory list and
> creates symbolic links for each file, where the name of the link is the hash value. (If the platform does not support symbolic 
> links, a copy is made.) This utility is useful as many programs that use OpenSSL require directories to be set up like this in 
> order to find certificates.

I'm not too sure, but to me it seems like those symlinks shouldn't be in the repo anyway (because at least they are not portable). Instead, c_rehash should be executed during build and it would either create symlinks, or named copies, depending on host platform.

Am I right?

PR #1005, opened 2020-11-27
Fix incorrect return value from SSLSocket_createContext()

### LINKED ISSUE #1020: SSL_CTX_set_alpn_protos call breaks SSLSocket_createContext return value
**Summary** 

The `SSL_CTX_set_alpn_protos` returns `0` on success and non-`0` on error, the opposite of the other `SSL_CTX_` calls, so its return code needs to be inverted before being used as the return value of `SSLSocket_createContext`.

**Details**

In **src/SSLSocket.c**, `SSLSocket_setSocketForSSL` calls `SSLSocket_createContext`, expecting success to result in a return value of `1`:

```
	if (net->ctx != NULL || (rc = SSLSocket_createContext(net, opts)) == 1)
	{
```

In `SSLSocket_createContext`, the various `SSL_CTX_` calls return `1` on success, so the `rc` variable is returned directly.  However, `SSL_CTX_set_alpn_protos` is different, it returns `0` on success (see [1] and [2]), and the `rc` checking in `SSLSocket_createContext` recognizes this:

```
	if (opts->protos != NULL && opts->protos_len > 0) {
        if ((rc = SSL_CTX_set_alpn_protos(net->ctx, opts->protos, opts->protos_len)) != 0)
        {
            if (opts->struct_version >= 3)
                SSLSocket_error("SSL_CTX_set_alpn_protos", NULL, net->socket, rc, opts->ssl_error_cb, opts->ssl_error_context);
            else
                SSLSocket_error("SSL_CTX_set_alpn_protos", NULL, net->socket, rc, NULL, NULL);
            goto free_ctx;
        }
	}
```
However, since this `SSL_CTX_` call is the last one in `SSLSocket_createContext`, the `rc` where `0` means success is used as the return value of `SSLSocket_createContext`, making `SSLSocket_setSocketForSSL` think that `SSLSocket_createContext` failed when `SSL_CTX_set_alpn_protos` in fact succeeded.  To be consistent with the other `SSL_CTX_` calls and with the use of `SSLSocket_createContext` in `SSLSocket_setSocketForSSL`, the `rc` value has to be inverted somehow.  For example, you can set `rc=0` before the `goto` used in the failure handling and set `rc=1` after the failure clause:

```
	if (opts->protos != NULL && opts->protos_len > 0) {
        if ((rc = SSL_CTX_set_alpn_protos(net->ctx, opts->protos, opts->protos_len)) != 0)
        {
            if (opts->struct_version >= 3)
                SSLSocket_error("SSL_CTX_set_alpn_protos", NULL, net->socket, rc, opts->ssl_error_cb, opts->ssl_error_context);
            else
                SSLSocket_error("SSL_CTX_set_alpn_protos", NULL, net->socket, rc, NULL, NULL);
            rc = 0;  /* <------------------------- */
            goto free_ctx;
        }
        rc = 1;  /* <------------------------- */
	}
```

(And maybe change those spaces to tabs, to match the coding convention used everywhere else in **SSLSocket.c**,.)

[1] https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_alpn_protos.html
[2] https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_alpn_protos.html

sasq64:
Fix incorrect return value from SSLSocket_createContext() using the `protos` field.

(I am not signing the ECA so I guess this PR is informative only)

### DIFF
--- a/src/SSLSocket.c
+++ b/src/SSLSocket.c
@@ -672,14 +672,18 @@ int SSLSocket_createContext(networkHandles* net, MQTTClient_SSLOptions* opts)
 
 #if (OPENSSL_VERSION_NUMBER >= 0x010002000) /* 1.0.2 and later */
 	if (opts->protos != NULL && opts->protos_len > 0) {
+        // SSL_CTX_set_alpn_protos() returns 0 on success as opposed to the other SSL functions,
+        // so we need to flip the meaning of rc or the return code will be wrong.
         if ((rc = SSL_CTX_set_alpn_protos(net->ctx, opts->protos, opts->protos_len)) != 0)
         {
+            rc = 0;
             if (opts->struct_version >= 3)
                 SSLSocket_error("SSL_CTX_set_alpn_protos", NULL, net->socket, rc, opts->ssl_error_cb, opts->ssl_error_context);
             else
                 SSLSocket_error("SSL_CTX_set_alpn_protos", NULL, net->socket, rc, NULL, NULL);
             goto free_ctx;
         }
+        rc = 1;
 	}
 #endif
 


a4z:
Hi @icraggs , what are you say to this PR? 

It seems @sasq64  is one of the user who do not want to have some steps like the eca in their PR , but it also seem I need exactly that fix for a current project.

can you pull that in, or would it help if I submit the same PR , since I have a eca ?

icraggs:
I'll add the changes.

a4z:
awesome, thanks a lot Ian!

icraggs:
Added in referenced commit.

PR #1011, opened 2020-12-05
Correct constness on multi-subscription functions

deonmcclung:
I recognize that this is a change to the API.

However, as the MQTTAsync_subscribeMany and MQTTAsync_unsubscribeMany functions do not actually change what these parameters are passing, it is correct to mark them as const, which allows for _much_ more flexibility on the part of the caller.

Even though this is a change of function signature, **the increase in constness should apply transparently for existing users**. I verified the following forms on Visual Studio and GCC compilers.

```
char* topic1[] = "topic1";
char* topic2[] = "topic2";
char* myTopics[] = {topic1, topic2};

int myQos[] = {0, 0};

MQTTAsync_subscribeMany(handle, 2, myTopics, myQos, respOpts);    // No problem
```

Even if the user was casting to make the old API work...
```
const char* myTopics[] =
{
   "topic1",
   "topic2"
};

int myQos[] = {0, 0};

MQTTAsync_subscribeMany(handle, 2, (char* const*)myTopics, myQos, respOpts);    // Nasty existing cast, but still no problem
```

Or, another...
```
const char* myTopics[] =
{
   "topic1",
   "topic2"
};

int myQos[] = {0, 0};

MQTTAsync_subscribeMany(handle, 2, (char**)myTopics, myQos, respOpts);    // Nasty existing cast, but still no problem
```

The following calling forms are now acceptable, whereas they weren't before.
```
const char topic1[] = "topic1";
char* topic2[] = "topic2";
const char* myTopics[] = {topic1, topic2};

const int myQos[] = {0, 0};

MQTTAsync_subscribeMany(handle, 2, myTopics, myQos, respOpts);    // Now, no problem
```

Also...

```
const char* myTopics[] =
{
   "topic1",
   "topic2"
};

int myQos[] = {0, 0};

MQTTAsync_subscribeMany(handle, 2, myTopics, myQos, respOpts);    // Now, no problem
```

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -957,7 +957,7 @@ exit:
 }
 
 
-int MQTTAsync_subscribeMany(MQTTAsync handle, int count, char* const* topic, int* qos, MQTTAsync_responseOptions* response)
+int MQTTAsync_subscribeMany(MQTTAsync handle, int count, const char* const* topic, const int* qos, MQTTAsync_responseOptions* response)
 {
 	MQTTAsyncs* m = handle;
 	int i = 0;
@@ -1078,13 +1078,13 @@ int MQTTAsync_subscribe(MQTTAsync handle, const char* topic, int qos, MQTTAsync_
 {
 	int rc = 0;
 	FUNC_ENTRY;
-	rc = MQTTAsync_subscribeMany(handle, 1, (char * const *)(&topic), &qos, response);
+	rc = MQTTAsync_subscribeMany(handle, 1, &topic, &qos, response);
 	FUNC_EXIT_RC(rc);
 	return rc;
 }
 
 
-int MQTTAsync_unsubscribeMany(MQTTAsync handle, int count, char* const* topic, MQTTAsync_responseOptions* response)
+int MQTTAsync_unsubscribeMany(MQTTAsync handle, int count, const char* const* topic, MQTTAsync_responseOptions* response)
 {
 	MQTTAsyncs* m = handle;
 	int i = 0;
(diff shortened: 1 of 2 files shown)


icraggs:
I don't want to change the API in such a way that will cause application programs to have to change.  To do that would necessitate bumping the major version number, which I'm not ready to do.  I think we added all the consts we could a while ago that were possible without afffecting applications.

deonmcclung:
My understanding is that this change would _not_ require the change of any existing application code. As I gave in the examples above, any way the user might have been using it would still compile fine, even if they were casting into the old parameter signature. The "No problem" comments are potential existing usages that I tried that still compile fine. I couldn't think of one that didn't work. The "Now, no problem" comments are things that the user could not do before, but are now acceptable by this change.

It is akin to a change like:

```
void func(int* arg);

//to

void func(const int* arg);
```

It is _technically_ a change to the API, but only in a way that the API promises not to change something where it didn't promise before. Old calls passing changeable memory just happily go along with the new form.

I very much respect your desire not to break any existing applications, but I don't think this would do it. If you can propose one calling form that may be in the userbase that would not compile with this, I would happily accept your decision to reject this for now.

Thanks Ian for looking at it.

icraggs:
There are compiler warnings as a result of the string const changes.  The int one looks like it could survive.

/home/travis/build/eclipse/paho.mqtt.c/test/test11.c:1146:39: warning: passing argument 3 of ‘MQTTAsync_unsubscribeMany’ from incompatible pointer type [-Wincompatible-pointer-types]

  rc = MQTTAsync_unsubscribeMany(c, 2, topics, &opts);

                                       ^~~~~~

In file included from /home/travis/build/eclipse/paho.mqtt.c/test/test11.c:27:0:

/home/travis/build/eclipse/paho.mqtt.c/src/MQTTAsync.h:1530:17: note: expected ‘const char * const*’ but argument is of type ‘char **’

 LIBMQTT_API int MQTTAsync_unsubscribeMany(MQTTAsync handle, int count, const char* const* topic, MQTTAsync_responseOptions* response);

deonmcclung:
Thanks for finding that. I guess I didn't have my warnings cranked up. Yeah, I agree with your decision. I'll revert the topic one and leave the qos one.

icraggs:
Ok, great.  I think this is good to go.

PR #1012, opened 2020-12-06
Fix for issue #1007

### LINKED ISSUE #1007: Automatic reconnect fails after long period disconnected
**Describe the bug**
Using the Async API to create a tcp ssl connection to a broker, enabling the auto reconnect feature and letting the system disconnected from the network for some hours (+10h) will prevent the reconnect function to work properly.
This does not happen (or I couldn't reproduce it) with plain tcp connections.

**To Reproduce**
To reproduce it: 
- Connect to an SSL broker. I've tested with AWS and test.mosquitto.org. 
- Enable the automatic reconnect option. Let it connect one time and then just pull out the network cable so that it keeps trying to reconnect.
- After enough time (for me it's been 10-12h) put the cable back in and it will never succeed reconnecting.

I've traced the problem down to MQTTAsync_connecting (see screenshots). There the parameter MQTTVersion is 0, producing that MQTTPacket_send fails without even sending anything over the wire. MQTTVersion should be "4" corresponding to version 3.1.1.
It shouldn't be 0, so somewhere there must be some memory overwrite? I'm not sure when or where it's been reset.

**Expected behavior**
It should always reconnect, no matter how much time it doesn't have connection.

**Screenshots**
MQTTVersion_0
MQTTPacket_send_connect
MQTTVersion_0_(2)

 - OS: Platform specific, but using the linux code of the library
 - Version 1.3.6 and OpenSSL 1.1.1g

**Additional context**
Destroying the client and creating a new one will solve the problem but then the automatic reconnect can't be trusted for long periods without connection.
This is a fragment of the log where you can see MQTTPacket_send_connect failing (because of the MQTTVersion being 0)
image

This one is the same system but at the first connection (succeeding)
image

alex-gimenez:
This PR intends to fix the behaviour described in https://github.com/eclipse/paho.mqtt.c/issues/1007
With this change, I can't reproduce the problem anymore.

Signed-off-by: Alejandro Giménez <alexx091@gmail.com>

### DIFF
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -2753,6 +2753,15 @@ static int MQTTAsync_connecting(MQTTAsyncs* m)
 #endif
 
 	FUNC_ENTRY;
+	if (m->c->MQTTVersion == MQTTVERSION_DEFAULT) //#1007
+	{
+		if (m->connect.details.conn.MQTTVersion == MQTTVERSION_DEFAULT)
+		{	
+			m->connect.details.conn.MQTTVersion = MQTTVERSION_3_1_1;
+		}
+	}
+	 else
+	 	m->connect.details.conn.MQTTVersion = m->c->MQTTVersion;
 	if (m->serverURIcount > 0)
 	{
 		serverURI = m->serverURIs[m->connect.details.conn.currentURI];


icraggs:
So this PR makes me uncomfortable - it feels like fixing a symptom not the source of the problem.  My suggested code would look like this:

	if (m->connect.details.conn.MQTTVersion == MQTTVERSION_DEFAULT) /* should not happen - #1007 */
	{
		Log(LOG_ERROR, -1, "MQTT version is 0 in MQTTAsync_connecting");
		m->connect.details.conn.MQTTVersion = (m->c->MQTTVersion == MQTTVERSION_DEFAULT) ? MQTTVERSION_3_1_1 : m->c->MQTTVersion;
	}

as the problem is m->connect.details.conn.MQTTVersion being set to 0.  If the cause is not a memory overwrite, but the field not being set correctly, then:

	if (m->connect.details.conn.MQTTVersion == MQTTVERSION_DEFAULT) /* should not happen - #1007 */
	{
		Log(LOG_ERROR, -1, "MQTT version is 0 in MQTTAsync_connecting");
		m->connect.details.conn.MQTTVersion = MQTTVERSION_3_1_1;
	}

Should be enough.  But we don't know at this point.

icraggs:
I've made a change with extra comments and the log message.

PR #1015, opened 2020-12-08
Properly uninitialize OpenSSL - segfault after calling ERR_free_strings()

### LINKED ISSUE #1014: OpenSSL Crash during MQTTClient_destroy
Looks like we have some issues during ERR_free_strings() in SSLSocket.c .. we are running openssl 1.0.2 .. do you think the issue is related to paho.mqtt.c or more to openssl? 

image

svermigo:
It looks that OpenSSL was not uninitialized properly and the wrong order of deinit is causing random segfaults. 
It seems like a similar problem was found on the PHP. 
See link for the PHP bug: [https://bugs.php.net/bug.php?id=72140](https://bugs.php.net/bug.php?id=72140)
The solution for PHP is [here](http://git.php.net/?p=php-src.git;a=blobdiff;f=ext/openssl/openssl.c;h=79f666acc5a6421eff70764446f5038e8c660907;hp=88e396c630976329e9931d1e37de1c2bb832ddd8;hb=05033c9ebd015baf825581e5725a594da277d560;hpb=6d3fa654b702c8762aa80ab795080f5c4464d677).
The correct shutdown procedure was found there [stackoverflow](https://stackoverflow.com/questions/29845527/how-to-properly-uninitialize-openssl/29927669#29927669)

Signed-off-by: Igor Šverma <igor@nibejpi.com>

### DIFF
--- a/src/SSLSocket.c
+++ b/src/SSLSocket.c
@@ -502,9 +502,9 @@ void SSLSocket_terminate(void)
 
 	if (handle_openssl_init)
 	{
-		EVP_cleanup();
-		ERR_free_strings();
 		CRYPTO_set_locking_callback(NULL);
+		ERR_free_strings();
+		EVP_cleanup();
 		if (sslLocks)
 		{
 			int i = 0;


icraggs:
That's not the same order of calls as the php fix, is it?  EVP_cleanup() comes first.

Apparently free_strings() doesn't do anything in OpenSSL version > 1.1, so it doesn't affect recent versions.

jumoog:
well i still use openssl 1.0.2 for legacy projects

svermigo:
> That's not the same order of calls as the php fix, is it? EVP_cleanup() comes first.
> 
> Apparently free_strings() doesn't do anything in OpenSSL version > 1.1, so it doesn't affect recent versions.

Yes, you are right. But based on messages on [stackoverflow](https://stackoverflow.com/questions/29845527/how-to-properly-uninitialize-openssl/29927669#29927669) or in the [OpenSSL mailing list](http://comments.gmane.org/gmane.comp.encryption.openssl.user/50784) it seems to be better in this way.

We are unfortunately maintaining a project with OpenSSL 1.0.2 as well.

icraggs:
It's fine to use OpenSSL 1.0.2.  My thought, apart from the fact that I used 1.0.2 in the past and wasn't aware of any problem with this ordering, is how do we know that this order is the best and will work on all the possible combinations of platforms and OpenSSL versions?  We don't, of course, so this is a bit of crossing fingers and hoping for the best, except for the combinations explicitly exercised in the CI tests.  If the original segfaults were random, have you run tests with the new ordering enough to be confident that they have now gone away?  

The fact that OpenSSL 1.1 and greater don't use these functions means we won't hopefully be affected by these factors in the future.

This is not to say not to accept this PR, but these are the thoughts going through my mind.

icraggs:
Thanks, the stackoverflow link looks like a good one. Should we add 

CRYPTO_set_id_callback(NULL);
ENGINE_cleanup();

as well?  I do remember vaguely removing some cleanup SSL code because it was interfering with an application's use of OpenSSL.  I'll be looking at OpenSSL issues in 1.3.9 so I'll probably be coming back to these questions then too.

svermigo:
I am testing it (unfortunately) only on one device and it looks good to me. I havent observed the crash for last day. I will monitor that, but I hope it solved my issue. 

Sorry, but I am not pretty sure about the CRYPTO_set_id_callback(NULL) and ENGINE_cleanup()

PR #1018, opened 2020-12-09
docs: fix simple typo, immmediately -> immediately

timgates42:
There is a small typo in src/Socket.c.

Should read `immediately` rather than `immmediately`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


icraggs:
Thanks.  I've added the change myself as the IP check didn't pass.

PR #1034, opened 2021-01-03
CMake improvements

Croydon:
CMake improvements:
  * Only build SHA1 and BASE64 tests if `PAHO_ENABLE_TESTING` is `ON`


Croydon:
I'm not going to provide my postal address in order to be able to sign the Contributor Agreement.

That said, feel free to do with these changes whatever you want.

So far, adding a single `if` and removing trailing white spaces is so small that it is not even protected by copyright

PR #1046, opened 2021-01-20
Fix issue with thread handles on Windows platform (close them after thread creation)

Foxhound31:
Some thread handles are left opened on Windows platform that is a kind of resource leak. This pull request contains the simplest solution - handles are closed just after thread creation. It's correct as handle doesn't used anywhere in library, and closing handle doesn't affect the thread itself.


icraggs (changes requested):
Thanks.  The variable declarations should be at the top of blocks to allow older compilers to work.  Hovever, as the return values friom Thread_start() aren't used, the CloseHandle call can be made within Thread_start(), then it only needs to be added once.

Foxhound31:
Thank you, I moved ClosedHandle() into Thread_start() as you suggested.

icraggs (approved):
Thanks!

PR #1051, opened 2021-01-22
Fix CMake targets install path

Roman3349:
This fix is needed to properly build [Debian package](https://tracker.debian.org/pkg/paho.mqtt.c).


icraggs:

PR #1073, opened 2021-03-05
if rc== SOCKET_ERROR , Can't call MQTTAsync_closeOnly

maikebing:
if rc== SOCKET_ERROR  , Can't  call MQTTAsync_closeOnly

### DIFF
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -1555,8 +1555,11 @@ static void nextOrClose(MQTTAsyncs* m, int rc, char* message)
 {
 	int was_connected = m->c->connected;
 	FUNC_ENTRY;
-
-	if (MQTTAsync_checkConn(&m->connect, m)  && rc!= SOCKET_ERROR)
+ 	if (rc == SOCKET_ERROR && m->c->net.socket>0)
+	{
+		m->c->net.socket = SOCKET_ERROR;
+	}
+	if (MQTTAsync_checkConn(&m->connect, m) )
 	{
 		MQTTAsync_queuedCommand* conn;
 
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -1556,7 +1556,7 @@ static void nextOrClose(MQTTAsyncs* m, int rc, char* message)
 	int was_connected = m->c->connected;
 	FUNC_ENTRY;
 
-	if (MQTTAsync_checkConn(&m->connect, m))
+	if (MQTTAsync_checkConn(&m->connect, m)  && rc!= SOCKET_ERROR)
 	{
 		MQTTAsync_queuedCommand* conn;
 


icraggs:
Do you have a test or scenario (trace?) that exhibits the problem you want to fix?

maikebing:
The test scenarios are hard to create!
When  communication has failed(rc =SOCKET_ERROR),  however  m->c->net.socket still greater than 0  ,   MQTTAsync_closeOnly 
  trying to send a message, so it break!

maikebing:
Because  rc  =SOCKET_ERROR , but  m->c->net.socket still greater than 0  !

icraggs (changes requested):
I don't think this is the right approach to the problem.   There are other cleanup operations that are avoided by setting the socket in this way and it's not obvious why it's being done.

Is it really a problem to attempt to send the disconnect packet?   To avoid this it would be better to set the connected flag to 0 if the rc is SOCKET_ERROR.

maikebing:
Although not the best solution, when the network quality is poor, it can avoid crashes, but there are still connections that are not released causing the number of connections to exceed the system limit.

icraggs:
Setting the socket variable to -1 skips the socket cleanup in MQTTAsync_closeOnly, leaving it more likely that sockets will be left around.  

Setting connected to 0 avoids the attempt to send disconnect but still runs the socket cleanup code (you said that your problem was the sending of a message, presumably the disconnect).

Even so, the attempt to send a disconnect shouldn't cause a problem - it might get a socket error, but no worse that that.  What are you seeing?

maikebing:
```
0xb6de1f88 in writev () from /lib/libc.so.6
#0  0xb6de1f88 in writev () from /lib/libc.so.6
#1  0x0043e1c0 in Socket_writev (socket=36, iovecs=0x87cfebf8, count=2, 
    bytes=0x87cfec20) at Socket.c:483
#2  0x0043e404 in Socket_putdatas (socket=36, buf0=0xadee95f8 "\340", 
    buf0len=2, bufs=...) at Socket.c:541
#3  0x0044b410 in WebSocket_putdatas (net=0x89f1e544, buf0=0x87cfecd0, 
    buf0len=0x87cfecd4, bufs=0x87cfecbc) at WebSocket.c:977
#4  0x00439c7c in MQTTPacket_send (net=0x89f1e544, header=..., buffer=0x0, 
    buflen=0, freeData=0, MQTTVersion=0) at MQTTPacket.c:224
#5  0x0043a720 in MQTTPacket_send_disconnect (client=0x89f1e528, 
    reason=MQTTREASONCODE_SUCCESS, props=0x0) at MQTTPacket.c:537
#6  0x004347ac in MQTTAsync_closeOnly (client=0x89f1e528, 
    reasonCode=MQTTREASONCODE_SUCCESS, props=0x0) at MQTTAsyncUtils.c:2339
#7  0x0043216c in nextOrClose (m=0x89f1e1f8, rc=-1, 
    message=0x50f904 "socket error") at MQTTAsyncUtils.c:1585
#8  0x0043350c in MQTTAsync_receiveThread (n=0x88d00498)
    at MQTTAsyncUtils.c:2002
#9  0xb6efc084 in start_thread () from /lib/libpthread.so.0
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
A debugging session is active.
```

maikebing:
```
int SocketConnected(int sock)
{
	if (sock <= 0)
		return 0;
	struct tcp_info info;
	int len = sizeof(info);
	getsockopt(sock, IPPROTO_TCP, TCP_INFO, &info, (socklen_t*)&len);
	if ((info.tcpi_state == TCP_ESTABLISHED)) {
		return 1;
	}
	else {
		return 0;
	}
}
```

maikebing:
```

static void MQTTAsync_closeOnly(Clients* client, enum MQTTReasonCodes reasonCode, MQTTProperties* props)
{
	FUNC_ENTRY;
	client->good = 0;
	client->ping_outstanding = 0;
	client->ping_due = 0;
	if (client->net.socket > 0)
	{
		MQTTProtocol_checkPendingWrites();

		if (client->connected && SocketConnected(client->net.socket) && Socket_noPendingWrites(client->net.socket))
			MQTTPacket_send_disconnect(client, reasonCode, props);
		Thread_lock_mutex(socket_mutex);
		WebSocket_close(&client->net, WebSocket_CLOSE_NORMAL, NULL);
```

maikebing:
if (client->connected && SocketConnected(client->net.socket) && Socket_noPendingWrites(client->net.socket))

PR #1076, opened 2021-03-11
Minor params names fix: new -> t_new, old -> t_old

inobelar:
## Proposal
Even if this is `C` library, maybe will be better to use **safer** names for params: `t_new` - instead of **`new`** (and symmetric `t_old` instead of `old`), which is reserved `C++` keyword.

## Warning fix
During built-in (not as library, but as directly compiled-into-binary) compilation, with `gcc`, for `c++`, compiler says: `MQTTTime.h:47:50: error: invalid parameter name: 'new' is a keyword`. This minor pull request fixes it.

### DIFF
--- a/src/MQTTTime.c
+++ b/src/MQTTTime.c
@@ -71,36 +71,36 @@ START_TIME_TYPE MQTTTime_now(void)
 
 #if defined(_WIN32) || defined(_WIN64)
 /*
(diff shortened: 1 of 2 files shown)


icraggs:
I've merged these changes into the develop branch.  Thanks.

PR #1081, opened 2021-03-17
Add generic proxy support

gambels:
Moved WebSocket proxy support to generic proxy support.

Signed-off-by: Sven Gambel <sven.gambel@gmx.net>

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -47,6 +47,7 @@ SET(common_src
   Base64.c
   SHA1.c
   WebSocket.c
+  Proxy.c
 )
 
 IF (NOT PAHO_HIGH_PERFORMANCE)
--- a/src/Clients.h
+++ b/src/Clients.h
@@ -145,8 +145,8 @@ typedef struct
 	void* context;                  /**< calling context - used when calling disconnect_internal */
 	int MQTTVersion;                /**< the version of MQTT being used, 3, 4 or 5 */
 	int sessionExpiry;              /**< MQTT 5 session expiry */
-	char* httpProxy;                /**< HTTP proxy for websockets */
-	char* httpsProxy;               /**< HTTPS proxy for websockets */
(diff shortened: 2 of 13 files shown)


beckenc:
Any chance to get this merged into develop?
This is very useful as it allows connection to a broker via a proxy connection - **not only for websockets but for generic connections.**

icraggs:
Hi.  Thanks.

1. There's an error in the build on one of the older Windows versions as you can see above - snprintf isn't redefined as it needs to be.  I think I sort of expected that that would be fixed before we went further.

2. I think the copyright header for Proxy.c should also credit the other authors of the original file as it's a direct copy of that code.

3. How do you suggest I test this?  Ideally it would be added to the CI suite.  Do you have a suggestion for a proxy to use locally for Windows, Linux and MacOS?

gambels:
Hello Ian,

the build on one of the older Windows version should be fixed now. Also the author of the original file was named in Proxy.c.

For testing purpose we used tinyproxy under Linux. This is a lightweight http(s) proxy daemon and should be also available for MacOS. Unfortunately I have no experience with a proxy server under Windows. Maybe the Squid proxy server is a solution. Howerver this is not as lightweight as the tinyproxy server.

I would suggest to set up one proxy server available for all build machines and integrate it into your CI suite. A dedicated broker, which is only reachable over the proxy, should be also installed. Afterwards you can set up a test case with a publish/subscribe message via the corresponding proxy.

Tinyproxy
 * https://github.com/tinyproxy/tinyproxy

gambels:
Hi Ian,

is there any news from your side, or are there open points regarding testability?

icraggs:
Thanks.  I'll think about adding a test later.  I'll merge this for now.

tardyp:
Thank you very much. 
I have verified this to work in my environment. proxy support in paho is something we have been looking forward for a while!

PR #1082, opened 2021-03-17
Skip connect and disconnect reordering.

gambels:
By adding commands in the command queue, make sure that the connect and disconnect command has always equal priority to prevent any reordering, which may lead to an unwanted disconnect.

Signed-off-by: Sven Gambel <sven.gambel@gmx.net>

### DIFF
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -758,7 +758,7 @@ static int MQTTAsync_unpersistInflightMessages(Clients* c)
 #endif
 
 /**
- * List callback function for comparing client handles and command types being CONNECT or DISCONNECT 
+ * List callback function for comparing client handles and command types being CONNECT or DISCONNECT
  * @param a first MQTTAsync_queuedCommand pointer
  * @param b second MQTTAsync_queuedCommand pointer
  * @return boolean indicating whether a and b are equal
@@ -778,7 +778,7 @@ static int clientCompareConnectCommand(void* a, void* b)
 		}
 	}
 	return 0;	//Item NOT found in the list
-}								   
+}
 
 
 int MQTTAsync_addCommand(MQTTAsync_queuedCommand* command, int command_size)
@@ -801,9 +801,18 @@ int MQTTAsync_addCommand(MQTTAsync_queuedCommand* command, int command_size)
 		if (head != NULL && head->client == command->client && head->command.type == command->command.type)
 			MQTTAsync_freeCommand(command); /* ignore duplicate connect or disconnect command */
 		else
-		{ 
+		{
 			ListRemoveItem(MQTTAsync_commands, command, clientCompareConnectCommand); /* remove command from the list if already there */
-			ListInsert(MQTTAsync_commands, command, command_size, MQTTAsync_commands->first); /* add to the head of the list */
+
+			/* do not change the order of consecutive CONNECT / DISCONNECT commands */
+			ListElement* current = NULL;
+			while (ListNextElement(MQTTAsync_commands, &current))
+			{
+				head = (MQTTAsync_queuedCommand*)(current->content);
+				if (head == NULL || !(head->command.type == CONNECT || head->command.type == DISCONNECT))
+					break;
+			}
+			ListInsert(MQTTAsync_commands, command, command_size, current); /* add to the head of the list */
 		}
 	}
 	else


src/MQTTAsyncUtils.c
```diff
@@ -801,9 +801,18 @@ int MQTTAsync_addCommand(MQTTAsync_queuedCommand* command, int command_size)
...
 			MQTTAsync_freeCommand(command); /* ignore duplicate connect or disconnect command */
 		else
-		{ 
+		{
 			ListRemoveItem(MQTTAsync_commands, command, clientCompareConnectCommand); /* remove command from the list if already there */
-			ListInsert(MQTTAsync_commands, command, command_size, MQTTAsync_commands->first); /* add to the head of the list */
+
+			/* do not change the order of consecutive CONNECT / DISCONNECT commands */
+			ListElement* current = NULL;
+			while (ListNextElement(MQTTAsync_commands, &current))
```
icraggs, line 809: So this is going to search through the command list from the beginning to find a connect or disconnect command which is at the end of the list, by the looks of it.  This could be expensive if the list is long.

icraggs (changes requested):
What is the situation you are trying to avoid?  What's the scenario - a test?  

It occurs to me that we should avoid having consecutive disconnect/connect commands and only have one of either, but only a test scenario could determine that.

src/MQTTAsyncUtils.c
beckenc: My understanding is, that in the current implementation CONNECT and DISCONNECT are always on the head of the list (line 806 in the original code). 
The search stops on the first command that is different than CONNECT / DISCONNECT. So that means. Max two iterations.

icraggs: Oh yes, we're going in that direction.  Should be ok. 

However, it seems like having a connect and disconnect next to each other at the head of the list maybe isn't a good idea.  

I guess what we're really trying to prevent is having a connect command come to the head of the list when there is already a disconnect there?

What's the sort of scenario you're trying to deal with?  Throughput, network conditions, OS?  I'd rather be able to have a test to include in the CI suite, otherwise I don't know if we've fixed anything or if future changes reintroduce a problem.

beckenc: > I guess what we're really trying to prevent is having a connect command come to the head of the list when there is already a disconnect there?

Yes, thats the point!

The scenario (the requirement) is that we are trying to renew the mqtt connection (resp. the underlying ssl session) every 48 hours. We're doing this by calling disconnect with a subsequent connect call.
**Without** that patch _CONNECT_ and _DISCONNECT_ are always **in the front of the list**.
That means for our scenario _CONNECT_ -> _DISCONNECT_ -> ....  (and in the end, we are disconnected).

**With** that patch, we still push _CONNECT_ and _DISCONNECT_ to the front of the list, but without reordering them. Means, if there is already a _DISCONNECT_ on the head, link _CONNECT_ to the second place in the list.
And that was the intention of the caller when calling disconnect with a subsequent connect call.
_DISCONNECT_ -> _CONNECT_

icraggs: Right, that's something I can test -- which I've just done.

This works for me without any changes if I wait for the disconnect call to complete (by waiting for the onSuccess callback).  
That's the way I intended it would work, and I'm tempted to "fix" the code so that any connect or disconnect command is ignored if there is an existing connect or disconnect at the head of the list.  Then you would have to do that.

Alternatively, it could be changed to allow a sequence of disconnect/connect at the head of the list, but no other sequences of connect and disconnect except for a single connect or disconnect.  I'm not so keen on this because I don't think it's necessary, and the code more complex, but it would be explicit.

beckenc: Got it! 

> That's the way I intended it would work, and I'm tempted to "fix" the code so that any connect or disconnect command is ignored if there is an existing connect or disconnect at the head of the list. Then you would have to do that.

That might be the best approach. Disconnect or Connect calls should further return error codes if they are dropped.
The behavior of the current implementation is somehow unexpected (reordering of connect / disconnect)

On the application side, I can synchronize subsequent calls with the callbacks.

> Alternatively, it could be changed to allow a sequence of disconnect/connect at the head of the list, but no other sequences of connect and disconnect except for a single connect or disconnect. I'm not so keen on this because I don't think it's necessary, and the code more complex, but it would be explicit.

That is what the patch is meant to be.

icraggs: Ok cool.  I'll put in a fix to ignore any connect/disconnect command if there is an existing one at the head of the list (it nearly does that anyway) and return a code which indicates this has been done.

beckenc: I've added some remarks to your https://github.com/eclipse/paho.mqtt.c/commit/da998eab0787bfba7882dc9d8b368b3e63459234#r50403206

icraggs:
Closing as this has been addressed as per the conversation above.

PR #1083, opened 2021-03-19
Add a ROS package manifest

Timple:
With a ROS package manifest, this package is buildable as-is in a Catkin/Colcon workspace. These tools are typically used to build ROS packages.

The manifest also gets installed into the directory so tools like rosdep and rospkg are able to detect it as well.

None if this should affect non-catkin/colcon builds.


icraggs:
OK, I was fearful this might lead to more maintenance work, but looks simple enough.

Timple:
Luckily it's still the pure CMake that is build. Only the package.xml is required to recognize it and to know the dependencies.

icraggs (commented):
The name package.xml seems pretty generic - I hope it's not going to affect another build.  

Could we add documentation of some sort?   Even just some comments in the package.xml file.

package.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+
+<!-- A package.xml in the root of a cmake projects allows it to be build using catkin (the ROS1 build system) -->
```
gavanderhoorn:
```suggestion
<!-- A package.xml in the root of a cmake projects allows it to be built using catkin (the ROS1 build system) -->
```

Timple:
It seems I cannot change the 'hardcoded' version number in the package.xml file ([source](https://answers.ros.org/question/377601/dynamic-version-in-packagexml/))

Let me know if you'r happy to take on the extra maintenance burden on this.

Timple:
As I read from your reference, this method is preferred since it would allow people to (easily) build from source.

But if this is not an option, injecting it with a patch would be the alternative route I had in mind :+1: .
I'll leave it up to the maintainers here, I wouldn't mind either way.

(I actually didn't know the `rosinstall_generator` could do this)

PR #1090, opened 2021-04-08
Make sure to initialize the mask on the Publish struct

### LINKED ISSUE #1079: Corrupt data being published over a websocket
**Describe the bug**
Using the latest `v1.3.8` release we noticed that data being published gets corrupted over time. Initially data is send just fine, but at a random point in time it starts sending data in a corrupted format and it will keep doing so until we fully stop and start our application again.

**To Reproduce**
Difficult to know what to describe here. We connect using a websocket and a proxy and we connect to AWS IoT Core (don't think that matters, but doesn't hurt to mention). We also run `v1.3.8` in a setup that doesn't use a websocket and a proxy (it uses MQTT over an SSL connection) and we didn't see the same issue in that setup. Hence the title mentioning the use of a websocket, yet we are not 100% sure these things are related.

**Expected behavior**
Data to be published and received without corruption.

Example of normal messages:
```
{"id":"NyFFWEUJhLWtuXsaXZpbMd","v":1082172151,"q":true,"t":1615804104165}
{"id":"4dwhRsu8SpaJ32oBe3GYie","v":1089564520,"q":true,"t":1615804104165}
{"id":"XiWCGL7cCAFahWwMteznd6","v":257496,"q":true,"t":1615804104165}
```

Example of corrupt messages:
```
{"9d":rNyFWEUhLW$uXs1XZp2Md"|"v"j108`718`70,rq":$rue|"t"j161e804c102g9}
+"idr:"44whR#u8S aJ3boBecGYi5","&":1a002b428d,"qr:tr%e,"$":1f158`431`279-
+"idr:"X9WCG7cCFahwMt5zndf","&":2f118`,"qr:tr%e,"$":1f158`431`279-
```

**Log files**
Please see this debug log: https://gist.github.com/svanharmelen/96188e34031e8fdee576c8889b8b9793 The first 3 publishes are as expected, the next 3 show the corruption.

**Environment (please complete the following information):**
 - OS: Linux
 - Version v1.3.8

svanharmelen:
After a lot of debugging we found this to be the root cause of the corruption reported in #1079

When not initializing the mask properly, the [check](https://github.com/eclipse/paho.mqtt.c/blob/master/src/WebSocket.c#L274) to see if a mask already exists might fail (depending on the existing data in that part of the memory) which then means that whatever is in the mask will be used for masking the data.

Fixes #1079

Signed-off-by: Sander van Harmelen <sander@vanharmelen.nl>

### DIFF
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -758,7 +758,7 @@ static int MQTTAsync_unpersistInflightMessages(Clients* c)
 #endif
 
 /**
- * List callback function for comparing client handles and command types being CONNECT or DISCONNECT 
+ * List callback function for comparing client handles and command types being CONNECT or DISCONNECT
  * @param a first MQTTAsync_queuedCommand pointer
  * @param b second MQTTAsync_queuedCommand pointer
  * @return boolean indicating whether a and b are equal
@@ -778,7 +778,7 @@ static int clientCompareConnectCommand(void* a, void* b)
 		}
 	}
 	return 0;	//Item NOT found in the list
-}								   
+}
 
 
 int MQTTAsync_addCommand(MQTTAsync_queuedCommand* command, int command_size)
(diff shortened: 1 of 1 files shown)


svanharmelen:
Let me know if I should make this PR against the `develop` branch instead... Yet it would be very nice if a bugfix release could be make for this one, as it's really causing a lot of issues in our setup. Thanks!

icraggs:
Wow!  Thanks for that debugging.  No need to recreate against develop, I can move the commit myself.  I'll put out a bugfix release - there are a couple of other fixes already lined up too.

svanharmelen:
Cool! I'll keep an eye on the releases than 👍🏻 

And next time I'll target `develop` to make things just a bit easier 😏

icraggs:
Merged into develop.

svanharmelen:
Any idea when we can expect 1.3.9?

PR #1096, opened 2021-05-02
Oss-fuzz integration

catenacyber:
Cf https://github.com/google/oss-fuzz/pull/5447 cc @icraggs


catenacyber:
friendly ping @icraggs

PR #1117, opened 2021-06-11
[README] Added description for CMake option PAHO_ENABLE_TESTING

inobelar:
Added (possibly missed) indicating that tests are enabled by default


icraggs:
Merged into develop.  Thanks.

Issue #60, opened 2016-02-04
Uninitialised variable in example program


jpwsutton:
migrated from Bugzilla [#427181](https://bugs.eclipse.org/bugs/show_bug.cgi?id=427181)
status RESOLVED severity _normal_ in component _MQTT-C_ for _---_
Reported in version _v0.5_ on platform _PC_
Assigned to: Ian Craggs

On 2014-02-01 02:22:07 -0500, Michael Andersen wrote:

> In the embedded C client for MQTT, the qos0pub.c example doesn't properly initialise the topic string, causing all sorts of mayhem.
> 
> The line in question: http://git.eclipse.org/c/paho/org.eclipse.paho.mqtt.embedded-c.git/tree/samples/qos0pub.c#n126
> 
> Ought to be:
> MQTTString topicString = MQTTString_initializer;
> 
> Thanks

On 2014-02-01 08:20:42 -0500, Ian Craggs wrote:

> Yep, initializing it would be a good idea.  By accident, it didn't cause a problem for me, which is why it wasn't initialized.  Fixed in master stream.
> 
> Thanks.

On 2014-08-21 09:42:17 -0400, Benjamin CabÃ© wrote:

> Please also fix on https://www.eclipse.org/paho/clients/c/embedded/ - I can confirm that this is causing indeed all sorts of mayhem :-)
> 
> Bonus points, Ian, if you fix it on your blog too :) http://modelbasedtesting.co.uk/?p=69

On 2014-08-21 17:03:46 -0400, Ian Craggs wrote:

> Sorry Benjamin!  Should be fixed in both places now

PR #1124, opened 2021-07-08
Fix MQTTAsync_terminate() to re-check the number of handles after re-acquisition of mqttasync_mutex

roscaviktor:
**Description**
Analyzing a few dumps revealed that the crash happens in the PAHO library in the background connection read thread, in function Socket_continueWrites(), which receives its socket argument pointing to de-allocated memory. Further investigation revealed that the cause for that is that the global variables, including variable mod_s in Socket.c, are cleaned up while the background threads are still alive.

Investigating further, the following race condition was found, leading to the possibility of having background threads running with global variables de-initialized:

1. MQTTAsync_destroy() , after freeing resources allocated to the given MQTTAsync handle, it checks if this was the last one, in which case it also calls MQTTAsync_terminate() to terminate the background threads and uninitialize the global variables

2. MQTTAsync_terminate() calls MQTTAsync_stop() which asks the background threads to terminate, by setting MQTTAsync_tostop to 1 (=true), and then waits for the background threads to actually terminate.

3. During that wait, mqttasync_mutex is released (and re-aquired afterwards), which opens the possibility that an MQTTAsync_create() and even MQTTAsync_connect() are called in the meantime

4. MQTTAsync_connect() resets  MQTTAsync_tostop  to 0 (false) and restarts the background threads if already terminated.

5. Even in that case events at 3-4 happen, MQTTAsync_terminate() continues by destroying the global variables, effectively corrupting the global data used by the background threads (and also by future API calls).

The main part of the fix is, in MQTTAsync_terminate(), to re-check the number of handles after re-aquisition of  mqttasync_mutex, and refrain from deleting the global variables if the number of handles is no longer zero.

**Call stack**
#0  0x00007fffdfb20d3b in Socket_continueWrites (pwset=0x7fffbfffebc0, sock=0x7fffbfffeba4)
    at /builds/di/stork/raven/Raven/src/tools/externals/paho.mqtt.c/src/Socket.c:1001
#1  0x00007fffdfb1e4c8 in Socket_getReadySocket (more_work=0, tp=0x7fffbfffece0, 
    mutex=0x7fffdfd4d3a0 <socket_mutex_store>, rc=0x7fffbfffeca8)
    at /builds/di/stork/raven/Raven/src/tools/externals/paho.mqtt.c/src/Socket.c:287
#2  0x00007fffdfb0ca72 in MQTTAsync_cycle (sock=0x7fffbfffedbc, timeout=1000, rc=0x7fffbfffedb8)
    at /builds/di/stork/raven/Raven/src/tools/externals/paho.mqtt.c/src/MQTTAsyncUtils.c:2854
#3  0x00007fffdfb08589 in MQTTAsync_receiveThread (n=0x7fffe406eea8)
    at /builds/di/stork/raven/Raven/src/tools/externals/paho.mqtt.c/src/MQTTAsyncUtils.c:1955
#4  0x00007ffff3ce56db in start_thread (arg=0x7fffbffff700) at pthread_create.c:463
#5  0x00007ffff2ecb71f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

**To Reproduce**
The crash is sporadically reproduced by running automated tests that are stressing MQTT test connection functionality in our application.

**Environment**
Docker Image: Linux 5.8.0-53-generic #60~20.04.1-Ubuntu SMP Thu May 6 09:52:46 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

Signed-off-by: Victor Rosca <rosca.viktor@gmail.com>

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -627,8 +627,6 @@ int MQTTAsync_connect(MQTTAsync handle, const MQTTAsync_connectOptions* options)
 	m->connect.context = options->context;
 	m->connectTimeout = options->connectTimeout;
 
-	MQTTAsync_tostop = 0;
-
 	/* don't lock async mutex if we are being called from a callback */
 	thread_id = Thread_getid();
 	if (thread_id != sendThread_id && thread_id != receiveThread_id)
@@ -636,6 +634,7 @@ int MQTTAsync_connect(MQTTAsync handle, const MQTTAsync_connectOptions* options)
 		MQTTAsync_lock_mutex(mqttasync_mutex);
 		locked = 1;
 	}
+	MQTTAsync_tostop = 0;
 	if (sendThread_state != STARTING && sendThread_state != RUNNING)
 	{
 		sendThread_state = STARTING;
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -198,7 +198,9 @@ void MQTTAsync_terminate(void)
 {
 	FUNC_ENTRY;
 	MQTTAsync_stop();
-	if (global_initialized)
+
+	/* don't destroy global data if a new client was created while waiting for background threads to terminate */
+	if (global_initialized && bstate->clients->count == 0)
 	{
 		ListElement* elem = NULL;
 		ListFree(bstate->clients);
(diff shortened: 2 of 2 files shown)


icraggs:
Thank you for the comprehensive description and fix.

PR #1131, opened 2021-07-27
Fix disconnect when there are pending writes

vm-vloz:
There is a bug that occurs with QoS1 or QoS2 transmissions.
Here's the sequence of events:
1. The MQTT client attempts to send a large message. That message is broken up into chunks. At some point the socket buffer fills up and an SSL_ERROR_WANT_WRITE occurs. This prompts the library to store the in-flight message (that was just interrupted) as a pending write.
2. The MQTT broker sends a (different) QoS1 or QoS2 message to the client. This occurs while the socket buffer is still full.
3. The MQTT client checks to see if there are any pending writes. If there are, this means the socket is full and is not ready to send any more data.
4. Because there are pending writes, the MQTT client treats that as a SOCKET_ERROR.
5. This error gets propagated up the call stack. Eventually the socket is closed as a result of the SOCKET_ERROR.
6. Closing the socket results in a disconnection. This triggers the LWT in the MQTT broker.
7. Because persistence is turned on, once the client automatically reconnects, it attempts to send the large message from step 1 again.
8. Also, because the MQTT broker has not had its message from step 2 acknowledged, the broker attempts to send this message again.
9. This repeats ad infinitum.

To solve this problem, I've added a queue of ACKs to the client. Once the pending writes are done the MQTT client will attempt to acknowledge all the messages in the queue. So, step 4 will not result in a SOCKET_ERROR, rather, it will result in the ACK message getting queued.

I've attached a log of the problem titled BeforeFix.log. The errors start on line 3209.
[BeforeFix.log](https://github.com/eclipse/paho.mqtt.c/files/6882172/BeforeFix.log)

Signed-off-by: vm-vloz <73361520+vm-vloz@users.noreply.github.com>

### DIFF
--- a/src/Clients.h
+++ b/src/Clients.h
@@ -137,6 +137,7 @@ typedef struct
 	List* inboundMsgs;              /**< inbound in flight messages */
 	List* outboundMsgs;				/**< outbound in flight messages */
 	List* messageQueue;             /**< inbound complete but undelivered messages */
+	List* outboundQueue;            /**< outbound queued messages */
 	unsigned int qentry_seqno;
 	void* phandle;                  /**< the persistence handle */
 	MQTTClient_persistence* persistence; /**< a persistence implementation */
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -356,6 +356,7 @@ int MQTTAsync_createWithOptions(MQTTAsync* handle, const char* serverURI, const 
 		bstate->clients = ListInitialize();
 		Socket_outInitialize();
 		Socket_setWriteCompleteCallback(MQTTAsync_writeComplete);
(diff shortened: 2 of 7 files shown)


icraggs:
Thanks for this.  I'll be looking at it soon.

nirmitdesai:
Hello,

Do we need to flush the outboundQueue in MQTTProtocol_retries method too? 
What if an ACK is queued but before it has chance to be sent over the wire, connection is lost and client is reconnected?
For QoS 1/2 topic, the broker would send the message again. 

I took the above commit and with modified MQTT client, we ran into a situation where we received duplicate messages on reconnection even when we had subscribed to the topic with QoS 2.

It's quite difficult to reproduce the issue but I thought I would get an initial discussion going.

Issue #1108, opened 2021-05-27
With MQTTCLIENT_PERSISTENCE_DEFAULT the paho code crashes for MQTTVERSION_5


bansalaj:
**Description**
If upon new connection or reconnection there are files in the persistence directory then the paho-mqttas code crashes.
On restoring commands from persistent directory it also restores the SUBSCRIBE commands from the file, and while writing to the file for MQTTVERSION_5 it does not store the qoss details. So after restoring commands it tries to process the command for SUBSCRIBE and as there is no qoss details the code crashes.

**To Reproduce**
Create Client with MQTTCLIENT_PERSISTENCE_DEFAULT specifying persistence directory
After Initial connection success
Subscribe to some topics
Terminate the process abruptly
On reconnect (with same client and to same broker) the code will crash.

**Expected behavior**
On termination of client either the persistent directory should be clear, if not able to clear than code should not crash.
For MQTTVERSION_5 qoss details should be added to the persistent file.

**Log files**
Attached crash.log
[crash.log](https://github.com/eclipse/paho.mqtt.c/files/6551531/crash.log)

** Environment (please complete the following information):**
 - OS: Windows 10
 - Version 1.3.8
 
**Additional context**

bansalaj:
Forgot to mention:
I was using regression test where i call connect <=>subscribe <=>disconnect <=> connect <=> subscribe <=> publish <=> disconnect (killing local broker) <=> reconnect.

But i was not calling MQTTAsync_Destroy() after every MQTTAsync_disconnect() call.

bansalaj:
What is the use of storing SUBSCRIBE message in the persistence file, and if we are storing then why not storing the full details?
To me it really doesn't make sense to store the SUBSCRIBE message, because on successful connection to the server, client can always send the SUBSCRIBE message to the server through success callback.

icraggs:
The subscribe is stored in the persistence file when it is put onto the command queue, which stores all commands before they are sent over TCP/IP.  Also, as a QoS 1 message, it could be retried on connection re-establishment.

bansalaj:
for the fix, if we add the qoss details support in the file while writing details of SUBSCRIBE command along with other respective changes like cleanup etc? will it suffice?

PR #1146, opened 2021-08-12
Fix for issue #1108 MQTTCLIENT_PERSISTENCE_DEFAULT crash

bansalaj:
…QTTVERSION_5 #1108

For MQTTVERSION_5 for SUBSCRIBE command qoss details was not added in persistent folder.
Also at time of retrieving for SUBSCRIBE & UNSUBSCRIBE command ptr value was not update properly for MQTTProperties value in the buffer.
Because of which MQTTProperties_read() would not return 1 and qcommand ends being free and later down the call stack crash would happen.

This fix covers above two issue.

Signed-off-by: ajit bansal ajitvkb1@hotmail.com


icraggs:
Thanks

PR #1149, opened 2021-08-23
Adding pkgconfig template file

nhorman:
Currently integrating paho with other projects is made a little harder
due to the lack of pkg-config files, as it prevents utilities like
autotools from using their built in macros to glean library install
switches and paths, especially when installed to non-standard locations.

Add the package config templates to allow use of pkg-config to determine
library install information

Signed-off-by: Neil Horman<nhorman@gmail.com>

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -25,6 +25,26 @@ CONFIGURE_FILE(VersionInfo.h.in
     @ONLY
     )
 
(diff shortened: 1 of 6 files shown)


nhorman:
ping, any feedback here?

nhorman:
ping, its been almost 6 months of this sitting here, is there any feedback?

PR #1032, opened 2021-01-02
Added optional PKCS#11 Support

cstaylor:
This will add optional support for the PKCS#11 interface with openssl. Users can leverage this to work with any HSM that supports the PKCS#11 interface.

Tests bring in SoftHSM2
Signed-off-by: Christopher Taylor cstaylor@argo.ai

PR #1157, opened 2021-10-01
Added optional PKCS#11 Support

argo-cstaylor:
This will add optional support for the PKCS#11 interface with openssl. Users can leverage this to work with any HSM that supports the PKCS#11 interface.

This replaces closed PR #1032 as the rules for who can issue a PR have changed.

Tests bring in SoftHSM2 Signed-off-by: Christopher Taylor cstaylor@argo.ai


argo-cstaylor:
There seems to be a problem with the `eclipsefdn/eca` since it is still referring to #1032 for this new PR.  Could an admin take a look and suggest a solution?

icraggs:
It looks like commit 5acc1956f43fbce675b34e303f3e20cadbc4f3c0 has an incorrect signed-off by tag:

Signed-off-by: Christopher Taylor cstaylor@argo.ai

It's missing the <> around the email address.  

And commit c6940b21e59f04f3f6b2433bbe1ad9dce6743a8e doesn't have a signed-off-by tag

pm1992:
Hello. Is this PR still being blocked by eclipsefdn/eca legal report? Or is that an error?

davisRoman:
Would be great if this PR could be merged. Is there anything I or the community can do to help move this along?

BeigeBadger:
What's the status of this PR? Can it be merged once the conflicts have been resolved?

bryant-ferguson:
I've been using this patch and it's useful, but there appears to be a problem that should be resolved before merging. (I see this is quite old, but at some point, maybe it will make it in.)

The approach to using Paho has always been that you can use common headers and choose between SSL or non-SSL libraries.

However, this PR puts a dependence on the "OPENSSL" and "PKCS11_HSM" compile definition into the *public headers* of Paho C. So, when you build against a library built with this patch enabled, you also need to set those two definitions.

I don't have a recommendation on how to proceed. Maybe see if you can remove the dependency from the public headers and avoid the problem all together?

KSroido:
hi, could anyone contacts the original auther, if they could aggree the eclipsefdn/eca  legal agreements, then we end-user could compile/use main release of paho and enjoy the PKCS11 feature
@argo-cstaylor 
@cstaylor

RoboAsystent:
I have tried paho_cs_pub with this patch included and tracing enabled on arm64. When using PKCS11 feature, it reports memory leaks which are HSM char* options. If I don't use them, then no leaks are reported. Check that if you gonna use this patch on your own.

PR #1167, opened 2021-10-26
Bugfix in paho_cs_pub.c and paho_cs_sub.c

Ashechol:
In the function **MQTTClient_connect()**, first variable **client** should be MQTTClient type, but in function **myconnect()** var client is a pointer of MQTTClient type. So I think it should be *client.
Signed-off-by: Ashechol <ashechol@qq.com>

### DIFF
--- a/src/samples/paho_cs_pub.c
+++ b/src/samples/paho_cs_pub.c
@@ -107,14 +107,14 @@ int myconnect(MQTTClient* client)
 		MQTTResponse response = MQTTResponse_initializer;
 
 		conn_opts.cleanstart = 1;
-		response = MQTTClient_connect5(client, &conn_opts, &props, &willProps);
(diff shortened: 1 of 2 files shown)


icraggs:
Thanks.  

I think the fix should rather be the parameter to myconnect should be of type MQTTClient, not MQTTClient*, similar to paho_c_pub.c.

This applies to paho_cs_sub.c too.

Ashechol:
Thanks for your advice，I have fixed my mistake.

litdemon:
It hasn't been fixed yet. In paho_cs_sub.c's main() function. myconnect() is called twice in different ways. myconnect(client) and myconnect(&client).

Ashechol:
Thanks a lot，I also fixed the myconnect function in paho_cs_sub.c

icraggs:
I've merged these changes into the develop branch.  Thanks.

PR #1174, opened 2021-11-10
Prevent calling `SSL_read` with `num == 0`

andrey-starodubtsev:
If whole packet was read from socket buffer, then `*actual_len == bytes`. But this means that `SSL_read` will be called with `num == 0`; in this case `ssl3_read_bytes` from OpenSSL, for example, return 0 and MQTT library treats connection as closed by remote peer.

Signed-off-by: Andrey Starodubtsev <andrey.starodubtsev@gmail.com>


icraggs:
Thanks.  Is this something you've seen happen, or spotted by inspection?

I guess that (*actual_len < bytes) should work also as we shouldn't exceed the bytes count?  Not that I want to change it, just checking.

PR #1181, opened 2021-12-03
Draft: avoid long delay in MQTTAsync_destroy #1107

### LINKED ISSUE #1107: 1sec delay in c library when MQTTAsync_destroy is called .
issue is that  MQTTAsync_destroy is taking 1 second due MQTTAsync_receiveThread which has select timeout of 1000L.(1sec)
Can we fix this issue, in our embedded device this is considerable delay,which is bringing latency between our microservices.
micoservice-a---->microservice b
we are using request reply pattern of paho c++ library
where we are getting a roundtrip delay of 1sec.[[Delay is from the c library as pointed above not in the transport]](since the communication is on same system we expect the delay to be less than 100ms )[[usbus and msgq gives 0.1millisecond latency]]

koalo:
This fixes #1107 by adding an additional pipe to the read file descriptors to enable writing to it when the select is waiting for a read. This unblocks the select.

There might be more elegant ways and I am open for other ideas or a completely other approach, but for me this at least avoids the 1 second delay of MQTTAsync_destroy for me. I have also reduced the sleep time in MQTTAsync_stop to further reduce the delay to 20 ms, but I am also OK with using the old sleep time of 100 ms.

Also it does not work for Windows yet.

### DIFF
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -2320,11 +2320,12 @@ static void MQTTAsync_stop(void)
 		{
 			int count = 0;
 			MQTTAsync_tostop = 1;
+			Socket_unblock();
 			while ((sendThread_state != STOPPED || receiveThread_state != STOPPED) && MQTTAsync_tostop != 0 && ++count < 100)
 			{
 				MQTTAsync_unlock_mutex(mqttasync_mutex);
 				Log(TRACE_MIN, -1, "sleeping");
-				MQTTAsync_sleep(100L);
+				MQTTAsync_sleep(20L);
 				MQTTAsync_lock_mutex(mqttasync_mutex);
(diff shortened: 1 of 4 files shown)


icraggs:
I've taken an alternative approach to this, which is now in the develop branch.  I've created an implementation which uses poll instead of select, as the poll system call can be interrupted when a socket is closed without having to have an extra pipe.  

Also, the use of poll addresses some other issues, so it seemed to make sense. If you try it out, let me know how it goes for you.  Thanks.

koalo:
Perfect, thanks a lot! I have already replaced my approach with yours in my system, but had no chance yet to fully verify it. However, looking at the code I am very confident that it solves the problem, so I will at least close this MR, because your approach is clearly better!

PR #1182, opened 2021-12-06
Add MQTTASyncSocketConnected to check sock  is TCP_ESTABLISHED

maikebing:
Check that the socket is valid and, if it is TCP_ESTABLISHED , that the connection is ok, otherwise the connection is no longer valid. 
The resources are released correctly in the event of poor network


icraggs (changes requested):
As you can see, the builds don't work on Windows or Mac

maikebing:
It's done.

jumoog:
your code is not properly formatted

maikebing:
I didn't find a format error, can you point it out? 
@jumoog

icraggs:
Ok, so what is the purpose here?  

You're calling the checkConnected function in checkDisconnect() and nextOrClose() which are both called when we are already disconnecting. One of the effects of calling it where you are would seem to be suppressing the calling of connectionLost() (because it's setting was_connected), which we don't want to do.

If the purpose is to detect TCP connection failures sooner, then I would have thought calling it in MQTTAsync_cycle() would be the right place. If it is to speed up the disconnect processing, then calling it in checkDisconnect could work, but to avoid continuing to wait if (m->c->outboundMsgs->count > 0). Both of these could be done.

The function should probably be in Socket.c along with all the other socket functions, rather than Client.c.

maikebing:
Thank you for your advice. I will try to modify and observe to solve the problem that the connection has actually been disconnected, but the status has not changed, so that the code still sends messages.   @icraggs

maikebing:
```
0xb6de1f88 in writev () from /lib/libc.so.6
#0  0xb6de1f88 in writev () from /lib/libc.so.6
#1  0x00449dd8 in Socket_writev (socket=52, iovecs=0xa6cfeb98, count=2, 
    bytes=0xa6cfebc0) at Socket.c:708
#2  0x0044a024 in Socket_putdatas (socket=52, buf0=0xadc46060 "\340", 
    buf0len=2, bufs=...) at Socket.c:766
#3  0x004573f8 in WebSocket_putdatas (net=0xa6d5475c, buf0=0xa6cfec70, 
    buf0len=0xa6cfec74, bufs=0xa6cfec5c) at WebSocket.c:972
#4  0x004457d8 in MQTTPacket_send (net=0xa6d5475c, header=..., buffer=0x0, 
    buflen=0, freeData=0, MQTTVersion=0) at MQTTPacket.c:224
#5  0x0044627c in MQTTPacket_send_disconnect (client=0xa6d54740, 
    reason=MQTTREASONCODE_SUCCESS, props=0xa5b4454c) at MQTTPacket.c:537
#6  0x0043ffa8 in MQTTAsync_closeOnly (client=0xa6d54740, 
    reasonCode=MQTTREASONCODE_SUCCESS, props=0xa5b4454c)
    at MQTTAsyncUtils.c:2358
#7  0x00440084 in MQTTAsync_closeSession (client=0xa6d54740, 
    reasonCode=MQTTREASONCODE_SUCCESS, props=0xa5b4454c)
    at MQTTAsyncUtils.c:2382
#8  0x0043ba58 in MQTTAsync_checkDisconnect (handle=0xa6d91860, 
    command=0xa5b44528) at MQTTAsyncUtils.c:946
#9  0x0043d394 in MQTTAsync_processCommand () at MQTTAsyncUtils.c:1492
#10 0x0043e138 in MQTTAsync_sendThread (n=0x0) at MQTTAsyncUtils.c:1776
#11 0xb6e8d084 in start_thread () from /lib/libpthread.so.0
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
A debugging session is active.

```

PR #1224, opened 2022-05-07
add ssl engine mode

guoyibiao0007:
Thank for your mqtt code.

i insert openssl engine mode for our own private key to load ,please check and merge it. tky!


icraggs:
Thanks, but as this doesn't pass the IP or CI check, I'll use #1247 instead.

PR #1231, opened 2022-05-31
Update of README with hints on using CMake for MQTT client apps.

fpagliughi:
Signed-off-by: fpagliughi <fpagliughi@mindspring.com>

I'm finally getting up to speed with CMake, and will hopefully PR some cleanup of the build systems for this library and the Paho C++ library as well.

In the meantime, I realized that the README could use some better information on how new users can start their own CMake projects that use the library. I also tried to make some of the existing examples in the README a little easier to peruse and updated some obsolete information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,22 +23,22 @@ Synchronous and various asynchronous programming models are supported.
 The Paho C client comprises four variant libraries, shared or static:
 
  * paho-mqtt3a - asynchronous (MQTTAsync)
- * paho-mqtt3as - asynchronous with SSL (MQTTAsync)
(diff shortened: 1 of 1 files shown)


fpagliughi:
Wait... how did this fail CI?!? I only changed the README :face_with_head_bandage:

jumoog:
sometimes the tests wont run

icraggs:
Thanks Frank. I've made some changes since this time which should make the tests more reliable. There were fixes both to the library and to the tests where edge cases could sometimes fail.

PR #1245, opened 2022-06-29
Update MQTTAsync.c

TonyC-code:
Fix for protos string becoming invalidated due to owner going out of scope.

Added MQTTStrdup() call to duplicate the protos string (as other strings are in the struct).


icraggs:
You'll need to check that 

    m->c->sslopts->protos

isn't NULL before calling MQTTStrdup, as is done in other cases.

TonyC-code:
Ok, thanks, done.

TonyC-code:
Hi. Is there any other actions I need to complete for this PR?

Thanks,

Tony.

PR #1247, opened 2022-07-12
ssl engine

shawshank119:
Thank you for your mqtt code. i add ssl engine mode for load private key. please check and merge.

### DIFF
--- a/src/MQTTAsync.h
+++ b/src/MQTTAsync.h
@@ -1062,6 +1062,7 @@ typedef struct
 	 * 2 means no ssl_error_context, ssl_error_cb
 	 * 3 means no ssl_psk_cb, ssl_psk_context, disableDefaultTrustStore
 	 * 4 means no protos, protos_len
+     * 5 means no ssl engine
 	 */
 	int struct_version;
 
@@ -1080,6 +1081,15 @@ typedef struct
 
 	/** The password to load the client's privateKey if encrypted. */
 	const char* privateKeyPassword;
+
+	/** Key mode "ENG" for engine or "PEM" for pem format Only used if struct_version is >= 6.*/
+	const char* keyType;
+
+    /** engineId for SSL Only used if struct_version is >= 6.*/
+    const char* engineId;
+
+    /** engine config file Only used if struct_version is >= 6.*/
+    const char* engineConfFile;
 
 	/**
 	* The list of cipher suites that the client will present to the server during the SSL handshake. For a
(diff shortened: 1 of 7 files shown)


shawshank119:
because "test95-3-offline-buffering-auto-reconnect-static" case always timeout , but i don't think is't my fault, so I'll take it out first.

src/MQTTAsync.h
```diff
@@ -1062,6 +1062,7 @@ typedef struct
 	 * 2 means no ssl_error_context, ssl_error_cb
 	 * 3 means no ssl_psk_cb, ssl_psk_context, disableDefaultTrustStore
 	 * 4 means no protos, protos_len
+     * 5 means no ssl engine
```
CIPop, line 1065: Nit : not aligned.

src/MQTTAsync.h
```diff
@@ -1081,6 +1082,15 @@ typedef struct
...
 	const char* privateKeyPassword;
 
+	/** Key mode "ENG" for engine or "PEM" for pem format Only used if struct_version is >= 6.*/
+	const char* keyType;
+
+    /** engineId for SSL Only used if struct_version is >= 6.*/
+    const char* engineId;
+
+    /** engine config file Only used if struct_version is >= 6.*/
+    const char* engineConfFile;
```
CIPop, line 1092: Nit: not aligned.

src/MQTTClient.h
```diff
@@ -686,6 +687,15 @@ typedef struct
 	/** The password to load the client's privateKey if encrypted. */
 	const char* privateKeyPassword;
 
+	/** Key mode "ENG" for engine or "PEM" for pem format Only used if struct_version is >= 6.*/
+	const char* keyType;
```
CIPop, line 691: Could this be an enum instead of string?

src/SSLSocket.c
```diff
@@ -608,27 +611,101 @@ int SSLSocket_createContext(networkHandles* net, MQTTClient_SSLOptions* opts)
...
+            ENGINE *eng = NULL;
+            EVP_PKEY *pkey = NULL;
+
+            if (opts->engineConfFile
+                && opts->engineId
+                && opts->privateKey)
+            {
+                if (!load_openssl_config) 
+                {
+                    if (CONF_modules_load_file(opts->engineConfFile, "openssl_conf", 0) != 1) {
```
CIPop, line 649: Why is loading configuration necessary? I would prefer that the default OpenSSL configuration is used instead of allowing applications modify it.
(E.g. https://github.com/dotnet/runtime/pull/88656/files#diff-b4a7a922ce304a412632e8b5a25c5f5a87fe5b3e63f639a5eb63a9c7b02668f9R279)

Either way, the code should allow for the case where `engineConfFile` is NULL.

CIPop (approved):
Thank you @shawshank119 !

The only part I don't like is not having the ability to use the default OpenSSL configuration (which in all my projects is default). It would be great if you could avoid loading the config (or load the default config) when `engineConfFile` is not set.

CIPop:
@shawshank119 I didn't see it originally: I think PRs should be sent to the `develop` branch first.

CIPop:
@shawshank119 I have previously written a test for our library using a Docker container, SoftHSM and p11:
https://github.com/Azure/azure-iot-sdk-c/blob/main/jenkins/openssl-engine/Dockerfile and a script that creates the simulated smart-cards: https://github.com/Azure/azure-iot-sdk-c/blob/main/jenkins/linux_openssl_engine.sh. 

We could re-use this to test the functionality end-to-end. 

@icraggs (and other maintainers) please let us know if the code change is acceptable and I can reserve some time to add the tests.

CIPop:
Parts of the code in this PR are identical with #1224.

icraggs:
> Parts of the code in this PR are identical with #1224.

@CIPop well #1224 doesn't pass the IP checks, so I'll base the changes on this PR instead of that one.

PR #1262, opened 2022-08-25
Add vcpkg installation instructions

JonLiu1993:
paho-mqtt is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for paho-mqtt and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build paho-mqtt, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/paho-mqtt/portfile.cmake). We try to keep the library maintained as close as possible to the original library. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,18 @@ In this case the libraries and executable are not linked against OpenSSL Librari
 apt-get install gcc-mingw-w64-x86-64 gcc-mingw-w64-i686
 ```
 
+## Building paho-mqtt - Using vcpkg
+
+You can download and install paho-mqtt using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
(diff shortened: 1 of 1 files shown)


jumoog:

To complete and submit a ECA, log into the Eclipse projects forge You will need to create an account with the Eclipse Foundation if you have not already done so. Be sure to use the same email address when you register for the account that you intend to use when you commit to Git. Go to https://accounts.eclipse.org/user/eca to sign the Eclipse ECA.

icraggs:
Merged into develop. Will be in the master branch soon.  Thanks!

PR #1270, opened 2022-09-21
Added URI protocol prefixes, 'mqtt://' and 'mqtts://'

fpagliughi:
A number of other libraries and tools use the `mqtt://` and `mqtts://` schemes for MQTT URI's, which makes it difficult to mix in applications using Paho C/C++/Rust, since you need to know which tools were written with which libraries to create a URI.

The `mqtt://` and `mqtts://` schemes also appears on the MQTT.org wiki: 
https://github.com/mqtt/mqtt.org/wiki/URI-Scheme

This PR **adds** these two schemes without changing existing behavior:
```
    mqtt://   <==>  tcp://
    mqtts://  <==>  ssl://
```

This PR does _not_, however:
- Address the web sockets, which could probably do something similar to CoAP, like `mqtt+ws://`, etc
- Implement username and password in the URI, which would be a nice addition.

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -328,9 +328,11 @@ int MQTTAsync_createWithOptions(MQTTAsync* handle, const char* serverURI, const 
 	if (strstr(serverURI, "://") != NULL)
 	{
 		if (strncmp(URI_TCP, serverURI, strlen(URI_TCP)) != 0
+		 && strncmp(URI_MQTT, serverURI, strlen(URI_MQTT)) != 0
 		 && strncmp(URI_WS, serverURI, strlen(URI_WS)) != 0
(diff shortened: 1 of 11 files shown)


icraggs (changes requested):
Thanks. Could we have some addition to the docs in MQTTClient.h and MQTTAsync.h and change a couple of tests to use the new prefixes?

fpagliughi:
Will do.

icraggs (approved):
Thanks very much Frank!

icraggs:
P.S. Would you like to open an issue for these?

- Address the web sockets, which could probably do something similar to CoAP, like mqtt+ws://, etc
- Implement username and password in the URI, which would be a nice addition.

PR #1289, opened 2022-11-10
memory leak

maikebing:
In any case, malloc should be free.


icraggs:
If we're going to do this, why not do all instances instead of just two?

Except for the one which we do not want to do this with - which is?

maikebing:
@icraggs 
Thanks for your advice, I fixed as many similar errors as possible. Please review the code.

maikebing:
CI can't pass, I can't see the problem, can you look at it?

Issue #1290, opened 2022-11-12
 Socket_getReadySocket    signal SIGSEGV, Segmentation fault.


maikebing:
The reason is not known at this time, and if there are any findings, feedback will continue 
```
Program received signal SIGSEGV, Segmentation fault.
[Switching to LWP 14087]
0xb6d9778c in memcpy () from /lib/libc.so.6
#0  0xb6d9778c in memcpy () from /lib/libc.so.6
#1  0x0046eaa4 in Socket_getReadySocket (more_work=0, timeout=1000, 
    mutex=0x5e6688 <socket_mutex_store>, rc=0xa36fed58)
    at paho.mqtt.c/src/Socket.c:528
#2  0x00466fe0 in MQTTAsync_cycle (sock=0xa36fee10, timeout=1000, 
    rc=0xa36fee14) at paho.mqtt.c/src/MQTTAsyncUtils.c:2995
#3  0x00464a84 in MQTTAsync_receiveThread (n=0xa3f1e5b8)
    at paho.mqtt.c/src/MQTTAsyncUtils.c:2016
#4  0xb6e8d084 in start_thread () from /lib/libpthread.so.0
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
A debugging session is active.
```

akai-123:
I encountered the same Segmentation fault, which often occurs when the mqtt server connection is not connected (for example, a non-existing ip is deliberately written). 
And found that when calling the Socket_getReadySocket function, mod_s.nfds in it will be modified. It seems that there are threads that are not locked by socket_mutex have changed mod_s.  Between the realloc and memcpy in Socket_getReadySocket,  mod_s.nfds seem may be modified by mod_s.nfds++ in the Socket_addSocket, resulting in a segmentation fault.

Stannieman:
We encountered at least 2 crashes that might be caused by or related to this.
```
0  libsystem_platform.dylib       0x16b2 _platform_strlen + 18
1  libsystem_c.dylib              0x458fc __vfprintf + 4639
2  libsystem_c.dylib              0x51fe4 _vsnprintf + 316
3  libsystem_c.dylib              0x520a4 vsnprintf + 95
4  AidEngine                      0x906d1 Log + 209
5  AidEngine                      0x9b0af myrealloc + 207
6  AidEngine                      0x8e8b5 Socket_getReadySocket + 245
7  AidEngine                      0xa1ede MQTTAsync_receiveThread + 446
[...]

PR #1291, opened 2022-11-13
Socket_getReadySocket signal SIGSEGV, Segmentation fault

maikebing:
Socket_getReadySocket signal SIGSEGV, Segmentation fault. #1290

### DIFF
--- a/src/Socket.c
+++ b/src/Socket.c
@@ -490,7 +490,7 @@ SOCKET Socket_getReadySocket(int more_work, int timeout, mutex_type mutex, int* 
 
 	FUNC_ENTRY;
 	Thread_lock_mutex(mutex);
-	if (mod_s.nfds == 0 || (mod_s.nfds == 0 && mod_s.saved.nfds == 0))
+	if (mod_s.nfds == 0 ||  mod_s.saved.nfds == 0)
 		goto exit;
 
 	if (more_work)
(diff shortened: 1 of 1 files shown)


src/Socket.c
```diff
@@ -490,7 +490,7 @@ SOCKET Socket_getReadySocket(int more_work, int timeout, mutex_type mutex, int*
 
 	FUNC_ENTRY;
 	Thread_lock_mutex(mutex);
-	if (mod_s.nfds == 0 && mod_s.saved.nfds == 0)
+	if (mod_s.nfds == 0 || (mod_s.nfds == 0 && mod_s.saved.nfds == 0))
```
icraggs: I don't think this is right. If mod_s.nfds == 0 then we won't get past the ||. If mod_s.nfds != 0 then both sides of the || will be false. So everything including the || and after is redundant.

src/Socket.c
```diff
@@ -521,14 +521,14 @@ SOCKET Socket_getReadySocket(int more_work, int timeout, mutex_type mutex, int*
 			else
 				mod_s.saved.fds_write = malloc(mod_s.nfds * sizeof(struct pollfd));
 		}
-		memcpy(mod_s.saved.fds_read, mod_s.fds_read, mod_s.nfds * sizeof(struct pollfd));
-		memcpy(mod_s.saved.fds_write, mod_s.fds_write, mod_s.nfds * sizeof(struct pollfd));
-
-		if (mod_s.saved.nfds == 0)
+		
+		if ( mod_s.fds_read == NULL || mod_s.fds_write == NULL)
```
icraggs, line 525: We definitely need to make sure the memory allocations worked, but then we'll need to return an error rather than just "no work to do". Probably PAHO_MEMORY_ERROR.

icraggs (changes requested):

src/Socket.c
maikebing: OK .

src/Socket.c
maikebing: sock = PAHO_MEMORY_ERROR;

maikebing:
CI can't pass, I can't see the problem, can you look at it?

PR #1305, opened 2022-12-05
checkEyecatchers() used to crash if a heap problem was detected

axelsommerfeldt:
checkEyecatchers() used vsnprintf() with a format specifier "%d" to print the wrong eyecatcher content, but an eyecatcher was defined as 'double'. Since "%d" expects an 'int' on the stack (which is usually 32 bit in size) but gets a 'double' instead (which is usually 64 bit), the following 'file' argument will be retrieved as wrong pointer value from stack, resulting in a crash in Log() -> vsnprintf() -> strlen().

This was fixed by defining 'eyecatcher' as 'uint64_t' (instead of 'double') and printing an eyecatcher using 'PRIx64' (instead of "d").

Signed-off-by: Axel Sommerfeldt <axel.sommerfeldt@fastmail.de>

### DIFF
--- a/src/Heap.c
+++ b/src/Heap.c
@@ -39,6 +39,7 @@ char* Broker_recordFFDC(char* symptoms);
 #include <string.h>
 #include <stdio.h>
 #include <stddef.h>
+#include <inttypes.h>
 
 #include "Heap.h"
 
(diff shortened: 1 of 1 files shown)


icraggs (approved):

icraggs:
Thank you

PR #1353, opened 2023-04-27
automatically load Windows Certificate Store

jumoog:
That's the equivalent to most Linux Distros ca-certificates packages.

### DIFF
--- a/src/SSLSocket.c
+++ b/src/SSLSocket.c
@@ -18,6 +18,7 @@
  *    Ian Craggs - fix for bug #480363, issue 13
  *    Ian Craggs - SNI support
  *    Ian Craggs - fix for issues #155, #160
+ *    Kilian von Pflugk - load Windows CA
  *******************************************************************************/
 
 /**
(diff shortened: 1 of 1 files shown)


CIPop:
Nitpick: ideally, an SChannel implementation instead of OpenSSL is required for Windows. 
While the fix is increasing security on Windows (TRAs are managed by Windows Update), there are several security features missing (e.g., verifying against the Untrusted Certificates store for WU revocation, maintaining the TLS stack up to date, etc.)

In general, OpenSSL should not be used on OSs that provide and maintain their own TLS stacks (Windows SChannel, macOS Secure Transport, etc.)

jumoog:
This code is used similarly in many other projects. The purpose is only to load the Windows certificates. Otherwise OpenSSL would not have any CA certificates.

> In general, OpenSSL should not be used on OSs that provide and maintain their own TLS stacks (Windows SChannel, macOS Secure Transport, etc.)

If you want you can create a new PR that replaces OpenSSL with Windows SChannel.

> there are several security features missing (e.g., verifying against the Untrusted Certificates store for WU revocation, maintaining the TLS stack up to date, etc.)

Feel free to add the missing security features.

CIPop (approved):
As noted in the comments: this PR increases the level of security for Paho on Windows as it relies on the OS-maintained CA store. I believe it should be merged in.

Please see my other comment regarding the overall level of security that can be achieved with OpenSSL on non-Linux systems.

Issue #465, opened 2018-05-29
MQTTV5: Enhanced authentication


icraggs:
#308 

There's a new AUTH packet, which can be sent and received by both clients and servers.  

1) Allow the clients to send an AUTH packet.

2) Allow the clients to respond to an incoming AUTH packet.

icraggs:
The auth packets can be exchanged before the connack is received, as well as afterwards for re-authentication.  Reading the spec, implementation of this is optional, so I'm going to defer the implementation of this until a follow on release.

PR #1363, opened 2023-05-26
Implement MQTTv5 Enhanced Authentication

ddassie-texa:
This PR adds MQTTv5 Enhanced Authentication to the library when using MQTTAsync.

The implementation on MQTTClient is currently incomplete and tests aren't yet provided, I will finish it once I receive feedback on the current implementation.

Implements: #465


ddassie-texa:
Implemented support for MQTTClient.

PR #1380, opened 2023-07-18
fix memory safty in tests

qzhuyan:
- buffer overflow
- use after free


icraggs:
This one also fixed in commit 08132d4cd76883b6e7c4a3d31891e7fa43171fbc

PR #1425, opened 2023-11-27
Bumped CMake minimum to 3.5 and added a CMake project version

fpagliughi:
While updating the Paho C++ client recently a few issues kept popping up. The C build is in pretty decent shape, but could use a few minor fixes and some modernization. This is the first of a few updates to the C build.

This PR contains two minor improvements.

1. Versions of CMake prior to v3.5 have been marked deprecated and are now generating annoying build warnings with recent versions.
``` 
CMake Deprecation Warning at CMakeLists.txt:20 (cmake_minimum_required):
Compatibility with CMake < 3.5 will be removed from a future version of CMake.
```
Version 3.5 should be good enough for a modern C project for a while going forward. Paho C++ has been using v3.5 for +5 years now, but will be moving to v3.11 in an upcoming version to support C++17.

2. Adds a CMake project version number. This also fixes a new warning message in recent versions of CMake that complain about the lack of a version. But more importantly, it lets other libraries and applications specify a minimum required version in their CMake builds. So, for example, the Paho C++ library can specify to require the latest C lib with:
```
find_package(eclipse-paho-mqtt-c 1.3.13 REQUIRE)
```
(This addition would finally allow me to fix and close the oldest Issue still active in the C++ library repo!)
 
I didn't remove the version files like `version.major`, etc, because they seem to also be used by the Makefile build. But to be safe I added a check to make sure the CMake project version and the file version match. If not, the build aborts with an error message.


fpagliughi:
Anyone know how I can see the details of the AppVeyor failure?

jumoog:
very strange it's not even in the [history](https://ci.appveyor.com/project/eclipsewebmaster/paho-mqtt-c/history)

fpagliughi:
Superseded by #1428

PR #1428, opened 2023-12-04
CMake cleanup, bump to v3.5, and submodule improvements

fpagliughi:
This is an overall cleanup of the CMake files with the desire to slightly modernize and make more usable as an inncluded subdirectory/submodule from other CMake projects.

This should be fully backward-compatible with the existing build other than requiring a newer version of CMake. No targets or exports were changed.

This supersedes #1425.

1. Versions of CMake prior to v3.5 have been marked deprecated and are now generating annoying build warnings with recent versions.
``` 
CMake Deprecation Warning at CMakeLists.txt:20 (cmake_minimum_required):
Compatibility with CMake < 3.5 will be removed from a future version of CMake.
```
Version 3.5 should be good enough for a modern C project for a while going forward. Paho C++ has been using v3.5 for +5 years now, but will be moving to v3.11 in an upcoming version to support C++17.

2. Adds a CMake project version number. This also fixes a new warning message in recent versions of CMake that complain about the lack of a version. But more importantly, it lets other libraries and applications specify a minimum required version in their CMake builds. So, for example, the Paho C++ library can specify to require the latest C lib with:
```
find_package(eclipse-paho-mqtt-c 1.3.13 REQUIRE)
```
(This addition would finally allow me to fix and close the oldest Issue still active in the C++ library repo!)
 
I didn't remove the version files like `version.major`, etc, because they seem to also be used by the Makefile build. But to be safe I added a check to make sure the CMake project version and the file version match. If not, the build aborts with an error message.

3. Added namespace aliases for all the libraries (like `eclipse-paho-mqtt-c::paho-mqtt3as`) so that the full namespaced target name can be used when including the project as a subdirectory. This is becoming a more common CMake pattern.
4. Reformat the CMake source files for consistency (indentation, capitalization, etc). Moved to lowerer-case function calls throughout which is more common lately and easier to read.


Aposhian:
Looks good to me!

Aposhian (approved):

jumoog (approved):

icraggs:
Yes, the whole point about the version.* files was that the version number only needed to be changed in one place for both builds, so the check will remind me.

PR #1432, opened 2023-12-14
Fix Crosscompilation for Windows on Linux

ernstblechaPT:
During compilation the build-script looks for libRpcRT4.dll, this file can not be found if the host system has a case-sensitve file system (e.g. Linux) and the case does not match (e.g. mingw32-64 provides librpcrt4.dll)


jumoog (approved):

ernstblechaPT:
Is there anything I can do to fix the failing AppVeyor build?

Aposhian:
I overlooked this PR, but added the same change here, along with some other fixes: #1453

icraggs:
Thank you. I've added the changes as a different commit because CMakeLists.txt had changed quite a bit and the merge was unnecessarily complicated. I've acknowledge the additional author in the commit message.

PR #1439, opened 2024-01-08
Fixed session present not be available in MQTT 5

Zephyrus29:
Session present does not function on MQTT 5, changed the check to be available for 3.1.1 and above rather than just 3.1.1


Zephyrus29:
Is there anyway to fix AppVeyor? Or I am reliant on a project admin to fix this

Aposhian:
Just noting here that this specifically fixes this for the synchronous client. The async client already correctly populates session present on any version, including MQTT v5.

icraggs:
Hi thanks. I did this as part of commit 2cb463b0e. I'm not sure why I put it in that commit. Thanks again.

Issue #1360, opened 2023-05-19
No way to receive failure data when reconnecting


ddassie-texa:
**Is your feature request related to a problem? Please describe.**
When using the reconnect system there is no way to receive failure data, this is because the `onFailure` callback gets reset to `NULL` when called for the first time, this means that when paho is trying to reconnect there is no way to know why the connect retry failed, getting this information would be very useful for the client to act upon (stop retrying if a certain failure happens while keep retrying in other cases), in my case the connect retry is failing because the broker has removed the client's username and password, but there is no way for the client to know that even tho the broker is sending a `CONNACK` with reason code `Not authorized`.

**Describe the solution you'd like**
I would like it if the `onFailure` callback were not reset or a similar system was provided so you could stop the app as it's not failing due to a network connection error but some other reason.

icraggs:
onSuccess/onFailure need to be only called once. Calling them again is not a solution. A new callback would be needed.

This is also raised in #1148

Issue #1148, opened 2021-08-16
Not able to know the connection failure reason in certain situation while using MQTT client with auto reconnect option


ashish-250:
**Scenario**
We attempt connecting the the server using MQTTAsync_connect with MQTTAsync_connectOptions autoreconnect. 
We get a successful connection after couple of retries. We get a onFailure callback set using MQTTAsync_connectOptions once, log the reason for failure and then get a connected callback set using MQTTAsync_setConnected API. Any future disconnections should be handled properly as we have configured auto-reconnect.
**Issue**
 Suppose, we disconnect after 12 hours and get a connectionLostCallback set using MQTTAsync_setCallbacks. The cause for the lost connection is always set to null as per the [docs](https://www.eclipse.org/paho/files/mqttdoc/MQTTAsync/html/_m_q_t_t_async_8h.html#a3900a98d7b1d58ad6e686bfce298bb6c). So, we do not know why we lost the connection. Now the client will attempt to reconnect and properly call the connected callback if succeeded but in case the connection attempt fails now with a different new reason, one would never get to know the reason outside of paho source code as paho does not call the onFailure and onSuccess callbacks more than once, refer #972. Someone may like to log the new failure reason for debugging purposes. Also, onFailure callback supplies MQTTAsync_failureData with a response code which can help us differentiate between a fatal and recoverable connection attempt failures. We do not get that information as well to make certain decisions. 
We also, cannot reinitiate a new MQTTAsync_connect with the fresh set of MQTTAsync_connectOptions as the reconnect logic puts a connect command on paho queue and we get MQTTASYNC_COMMAND_IGNORED for our manual attempt. 

I tried looking up any alternative apis for setting onFailure callback [here](https://www.eclipse.org/paho/files/mqttdoc/MQTTAsync/html/_m_q_t_t_async_8h.html), but had no luck. 

Is there anything, you can suggest to get the onFailure callback when the client disconnect after a certain interval of time and then fails to connect back?

icraggs:
There is always the option of reconnecting entirely by using the connectionLost callback and then calling connect when you are ready and with the options you choose.

Once you are in a series of reconnect attempts, the updateConnectOptions callback can be used to modify username and password.

If you want to try an alternative host/port to connect to, then you can use the serverURIs connect option.

[...]

PR #1442, opened 2024-01-22
Add Connection error callback to fix issue 1148

Ivan-Bolshakov:
Added a callback message with an error connecting to the broker. This message is sent during all connection and reconnection attempts (including automatic reconnection attempts).
This MR solves problems https://github.com/eclipse/paho.mqtt.c/issues/1148 and https://github.com/eclipse/paho.mqtt.c/issues/1360

Signed-off-by: Ivan Bolshakov <ivanvaxi@gmail.com>

### DIFF
--- a/src/MQTTAsync.h
+++ b/src/MQTTAsync.h
@@ -430,6 +430,25 @@ typedef void MQTTAsync_connectionLost(void* context, char* cause);
 typedef void MQTTAsync_connected(void* context, char* cause);
 
 /**
+ * This is a callback function, which will be called when the client library
+ * have some errors during connection process. This is superfluous when the
+ * connection is made in response to a MQTTAsync_connect call, because the
+ * onFailure callback can be used.  It is intended for use when automatic
+ * reconnect is enabled or after MQTTAsync_reconnect() function, so that when
+ * a connection attempt failed in the background, the application is notified
+ * and can take any required actions.
+ *
+ * <b>Note:</b> Neither MQTTAsync_create() nor MQTTAsync_destroy() should be
+ * called within this callback.
+ * @param context A pointer to any application-specific context. The
+ * the <i>context</i> pointer is passed to each of the callback functions to
+ * provide access to the context information in the callback.
+ * @param code A numeric code identifying the error.
+ * @param message Optional text explaining the error. Can be NULL.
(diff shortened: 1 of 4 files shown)


Aposhian:
Looks good to me!

Aposhian:
Do we want to consider adding this to `MQTTAsync_setCallbacks`? Or do we want to not do that in order to not change ABI compatibility?

src/MQTTAsync.h
```diff
@@ -429,6 +429,25 @@ typedef void MQTTAsync_connectionLost(void* context, char* cause);
...
+ *
+ * <b>Note:</b> Neither MQTTAsync_create() nor MQTTAsync_destroy() should be
+ * called within this callback.
+ * @param context A pointer to any application-specific context. The
+ * the <i>context</i> pointer is passed to each of the callback functions to
+ * provide access to the context information in the callback.
+ * @param code A numeric code identifying the error.
+ * @param message Optional text explaining the error. Can be NULL.
+ */
+typedef void MQTTAsync_connectionError(void* context, int code, const char *message);
```
Aposhian: I would recommend that we rename `message` to `cause` to be consistent with the other callback signatures. Also, it is worth noting that `const char *` is used here while other callbacks use `char *`. I think that is ok, but it is a difference.

Aposhian:
I have tested this and confirmed it to work. I created initial support on the C++ client for using this API: https://github.com/fireflyautomatix/paho.mqtt.cpp/tree/connection-error-callback

src/MQTTAsync.h
Ivan-Bolshakov: I am glad that my changes were useful to you! Thanks for the comments. Corrected MQTTAsync_connectionError description

Aposhian (approved):

jumoog (approved):

Ivan-Bolshakov:
@Aposhian @jumoog Thank you for your review! 
@icraggs Could you please watch this MR when you have time?

icraggs:
I don't think we want to add it to set callbacks, as that would cause some backward incompatibilities which we could manage, but is more complicated. As it is, as it is an interface change, I think it has to go into the 1.4 stream.

PR #1450, opened 2024-02-15
Add cause to connection lost callback

Aposhian:
I've noticed that the connection lost callback gets a string for the cause, but it is always set to null. It looks like upstream, many of the times, a cause is given, and I would like to propagate that.

This might not be the exactly correct approach, since I don't exactly understand the logic of the `nextOrClose` callback.


jumoog (approved):

Aposhian:
I haven't followed up on this in quite some time now, but I have been running with this change for over a year now with no issues. Can we consider getting this merged?

Issue #1426, opened 2023-11-27
Solution for #1383 causes build to fail on LibreSSL (missing OPENSSL_thread_stop())


nickpelone:
**Describe the bug**
I noticed when I updated the Rust version of this library it pulled in an update to the underlying `-sys` crate and thus this library(also mentioned in this release https://github.com/eclipse/paho.mqtt.rust/releases/tag/v0.12.3), which then stopped building correctly against LibreSSL:

```log
  = note: ld: error: undefined symbol: OPENSSL_thread_stop                                                                                                                                                                                                            
          >>> referenced by MQTTAsyncUtils.c:1827 (/home/queen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/paho-mqtt-sys-0.9.0/paho.mqtt.c/src/MQTTAsyncUtils.c:1827)                                                                                         
          >>>               MQTTAsyncUtils.c.o:(MQTTAsync_sendThread) in archive /home/queen/ccm-workspace/ccm-shared/target/debug/deps/libpaho_mqtt_sys-a95dc17b816e9955.rlib                                                                                         
          >>> referenced by MQTTAsyncUtils.c:2336 (/home/queen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/paho-mqtt-sys-0.9.0/paho.mqtt.c/src/MQTTAsyncUtils.c:2336)                                                                                         
          >>>               MQTTAsyncUtils.c.o:(MQTTAsync_receiveThread) in archive /home/queen/ccm-workspace/ccm-shared/target/debug/deps/libpaho_mqtt_sys-a95dc17b816e9955.rlib                                                                                      
          cc: error: linker command failed with exit code 1 (use -v to see invocation)                                                                                                                                                                                
          MQTTAsyncUtils.c:142 (/home/queen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/paho-mqtt-sys-0.9.0/paho.mqtt.c/src/MQTTAsyncUtils.c:142)(MQTTAsyncUtils.c.o:(MQTTAsync_randomJitter) in archive /home/queen/ccm-workspace/ccm-shared/target/debug/deps/libpaho_mqtt_sys-a95dc17b816e9
955.rlib): warning: rand() may return deterministic values, is that what you want?                                                                                                                                                                                    
[...]

Issue #1427, opened 2023-12-02
Some improvements and cleanup of CMake build


fpagliughi:
I've started working on some updates to the CMake build with the following goals:

1. Upgrade the targeted version of CMake to v3.5. It seems that earlier versions are being deprecated and causing build warnings with current versions of CMake. (See  #1425)
2. Set the CMake version of the project so that downstream users can request/check a minimum version in their CMake builds. (See #1425)
3. Make it easier to use Paho C as a submodule in other projects. A few minor things - like aliasing the namespace - would make it easy to include the whole project as a subdirectory in CMake.
4. Do a general cleanup of the CMake files to make them more consistent and readable.
5. Everything should be backward compatible - no breaking changes.

I'm interested in any other ideas people might have to make this easier to use in their projects, especially if anyone is maintaining the this code in a package manager and needs to add any work-arounnds or patches to get things compiling.

icraggs:
Thanks. The PR has been merged into the develop and 1.4 branches.

PR #1451, opened 2024-02-16
Support build against LibreSSL

marcfir:
Some projects use LibreSSL instead of OpenSSL.
In [umati/Dashboard-OPCUA-Client](https://github.com/umati/Dashboard-OPCUA-Client) we use a patch for this, but direct support would simplify our build https://github.com/umati/Dashboard-OPCUA-Client#633.

This also solves #1426


jumoog (approved):

GoetzGoerisch:
@icraggs anything we can help to get this merged?

icraggs:
My main question here is how does this get tested? What OSes would be supported? At the moment the CI tests are run for MacOS, Linux and Windows. Installation of LibreSSL in a Linux environment, at least on my usual Ubuntu OS, seems like it's not a readily installable package and it's not available by default in the CI environments I've looked at so far. I could merge it as unsupported but that would not be ideal.

GoetzGoerisch:
Thanks for the explanation.
So adding a workflow which builds and tests the compile option would be sufficient?

icraggs:
That would be a great help, yes. I also have to check the CMake changes against PR #1427 which I haven't merged yet, but intend to do so.

marcfir:
@icraggs I added tests for Linux and MacOS. The tests on Windows go into a timeout [CI LibreSSL](https://github.com/marcfir/paho.mqtt.c/actions/runs/8421189807). But tests with OpenSSL do the same [CI OpenSSL](https://github.com/marcfir/paho.mqtt.c/actions/runs/8421189800).

GoetzGoerisch:
@icraggs anything missing to get this merged?

icraggs:
@GoetzGoerisch I've been in North America for a week for the eclipse. I'm back now, and just been merging Frank's PR for CMake into the 1.4 branch. This PR is next on my list.

src/CMakeLists.txt
```diff
@@ -189,15 +193,28 @@ ENDIF()
 INSTALL(FILES MQTTAsync.h MQTTClient.h MQTTClientPersistence.h MQTTProperties.h MQTTReasonCodes.h MQTTSubscribeOpts.h MQTTExportDeclarations.h
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
-IF (PAHO_WITH_SSL)
-    SET(OPENSSL_ROOT_DIR "" CACHE PATH "Directory containing OpenSSL libraries and includes")
-    find_package(OpenSSL REQUIRED)
+IF (PAHO_WITH_SSL OR PAHO_WITH_LIBRESSL)
+    
+    IF (PAHO_WITH_LIBRESSL)
+        SET(OPENSSL_ROOT_DIR "" CACHE PATH "Directory containing LibreSSL libraries and includes")
```
icraggs: In merging this PR, I don't understand this line. Are we zeroing out the LIBRESSL root_dir, in which case the first parameter should be LIBRESSL_ROOT_DIR?

src/CMakeLists.txt
```diff
@@ -189,15 +193,28 @@ ENDIF()
...
+        find_package(LibreSSL REQUIRED)
+        SET(SSL_INCLUDE_DIR ${LIBRESSL_INCLUDE_DIR} CACHE PATH "Directory containing SSL includes")
+        SET(SSL_LIBRARY_NAME LibreSSL CACHE STRING "Name of the used SSL library")
+    ELSE()
+        SET(LIBRESSL_ROOT_DIR "" CACHE PATH "Directory containing OpenSSL libraries and includes")
```
icraggs: Same question as above but in reverse

src/CMakeLists.txt
```diff
@@ -189,15 +193,28 @@ ENDIF()
...
+        find_package(OpenSSL REQUIRED)
+        SET(SSL_INCLUDE_DIR ${OPENSSL_INCLUDE_DIR} CACHE PATH "Directory containing SSL includes")
+        SET(SSL_LIBRARY_NAME OpenSSL CACHE STRING "Name of the used SSL library")
+    ENDIF()
+    message("Use ${SSL_LIBRARY_NAME} at ${LIBRESSL_ROOT_DIR}")
```
icraggs: This doesn't seem right for OPENSSL, is it?

icraggs (commented):
3 questions as a result of merging

src/CMakeLists.txt
marcfir: Yes, these lines are mixed up

src/CMakeLists.txt
marcfir: Yes. I add a fix for that.

icraggs:
Ok, I merged the PR yesterday, with a couple of changes of my own to answer my questions, into the develop and 1.4 branches. You can check my changes there. 

If you propose any further changes they should be targeted at the develop branch, not master, as the CMake build files have changed due to another PR I've merged.

icraggs:
I'm going to close this PR now, as it's in the develop/1.4 branches. If you have any further comments you can make them here or in another PR/issue. Thanks for your contributions.

fpagliughi:
Apologies, I hadn't seen this PR before it was merged. I just wanted to ask a few questions before this is released. @marcfir 

I started writing this message with a number of questions about the use of cache variables where regular variables would have been adequate (preferable?) and why the `xxxSSL_ROOT_DIR` variables were being cleared each run, as if it were expected that the `find_package()` call would be setting this...

But then I realized this behavior was already in the CMake files from before the PR!
```
IF (PAHO_WITH_SSL)
    SET(OPENSSL_ROOT_DIR "" CACHE PATH "Directory containing OpenSSL libraries and includes")
    find_package(OpenSSL REQUIRED)
```

So I guess the style was just copy-pasted for this PR?

This (the original usage) doesn't seem correct to me. I don't think the CMake file should be randomly clearing a cache variable. The expectation (the purpose of the cache) is to retain variables between runs. Also, the proper usage for `OPENSSL/LIBRESSL_ROOT_DIR` is for the user to pass _in_ a hint (as an environment variable) about where the library can be found.

_Normally_ this isn't done and the variable is blank, which also makes the current configuration output look broken:
```
-- Found OpenSSL: /usr/lib/x86_64-linux-gnu/libcrypto.so (found version "3.0.2")  
Use OpenSSL at                                       <--- *** At nowhere? Looks like a bug ***
SSL_INCLUDE_DIR: /usr/include  
```
Then, when dumping the library information, it would be good to show the include directory and the main SSL library file (with path). That would have helped me recently on Windows where it was finding the headers from and older version (1.1) and then linking to a DLL with a different version (3.x), resulting in missing symbols in the link.

I suggest changing the `src/CMakeLists.txt` to this:
```
if(PAHO_WITH_SSL OR PAHO_WITH_LIBRESSL)
  if(PAHO_WITH_LIBRESSL)
    find_package(LibreSSL REQUIRED)
    set(SSL_LIBRARY_NAME LibreSSL)
    message(STATUS "Using LibreSSL with headers at ${LIBRESSL_INCLUDE_DIR}, lib ${LIBRESSL_SSL_LIBRARY}")
  else()
    find_package(OpenSSL REQUIRED)
    set(SSL_LIBRARY_NAME OpenSSL)
    message(STATUS "Using OpenSSL with headers at ${OPENSSL_INCLUDE_DIR}, lib ${OPENSSL_SSL_LIBRARY}")
  endif()
  ...
```

I can send a PR.

marcfir:
Hi @fpagliughi,
Yes, I just copied the style of including OpenSSL as is. Your suggestion and PR looks very good.

PR #1457, opened 2024-03-05
Fix timeout of send thread in Async client for v5

TiejunZhou-CN:
When the outgoing packets fails to be sent due to hitting the limitation of the inflight maximum, it will wait on send_cond/send_sem. Send thread should be woken up on PUBACK message received. So it does not need to wait until timeout (1 second).

### DIFF
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -3078,7 +3078,15 @@ static MQTTPacket* MQTTAsync_cycle(SOCKET* sock, unsigned long timeout, int* rc)
 				else if (msgtype == PUBREC)
 					*rc = MQTTProtocol_handlePubrecs(pack, *sock, &pubToRemove);
 				else if (msgtype == PUBACK)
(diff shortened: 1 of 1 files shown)


icraggs:
If this solution is correct (I want to check), then it will apply to PUBCOMP too, in case the QoS of the message being sent is 2.

TiejunZhou-CN:
Thanks for your suggestion! I run the test, and the same issue exists in QoS2 too. The test is modified from sample MQTTAsync_publish.c and attached. To run it, simply start a mosquitto broker in localhost. Before fix, you will see random stuck for 1 second. It does not always reproducible since it is a race condition. Run multiple times and the issue will eventually be reproduced.
[MQTTAsync_publish.zip](https://github.com/eclipse/paho.mqtt.c/files/14503158/MQTTAsync_publish.zip)

icraggs:
I've merged this fix into the develop and 1.4 branches. Thanks!

Issue #1430, opened 2023-12-11
long connection latency when setting up multiple mqtt connections


zhen9910:
Description:
    When setting up multiple connections using paho.mqtt.c library, the first connection is able to be set up very fast, about 200ms; but all the other connections are connected very slow, about 2s.

   I first experienced this performance issue using paho.mqtt.rust v 0.12. https://github.com/eclipse/paho.mqtt.rust/issues/217
As paho.mqtt.rust is a wrapper around paho.mqtt.c library, I can also reproduce this same perforamance issue using paho.mqtt.c.

   So I want to ask if this is a known constraint or something that can be improved?

Reproduce:
  This can be reproduced using attached sample code with paho-mqtt3a on Linux ubuntu 20.04.
[mqtt_sample.txt](https://github.com/eclipse/paho.mqtt.c/files/13640043/mqtt_sample.txt)

Log:
```
./build/mqtt_sample
====== Creating MQTT client ======
====== creating MQTT connect ======
====== Successful connection ====== 
====== Connect Time elpased is 204 ms ======

====== Creating MQTT client ======
====== creating MQTT connect ======
====== Successful connection ====== 
====== Connect Time elpased is 2105 ms ======

====== Creating MQTT client ======
====== creating MQTT connect ======
====== Successful connection ====== 
====== Connect Time elpased is 2106 ms ======
```
Trace:
 Attached in trace.txt

[trace.txt](https://github.com/eclipse/paho.mqtt.c/files/13640041/trace.txt)

icraggs:
I realise this is a long time later, but thanks for pointing this out.

PR #1458, opened 2024-03-08
Reduce the delay of ASYNC connection

TiejunZhou-CN:
Resolve the issue mentioned in #1430 

Move the MQTT layer connect from receive thread to send thread as the TCP/Websocket connection is done in send thread. The root cause of the 2 seconds delay mentioned in issue is that, MQTT connection needs to wait two rounds of timeout (1s) polling a socket to be ready. For the first time, `Socket_getReadySocket` is polling events for first socket before second socket is connected and added to fds_read. For the second time, the second socket is connected, but there is no incoming data. By moving the `MQTTAsync_connecting` to send thread, MQTT connection can start immediately after underlayer connection is established.

I also removed 100ms sleep after checking `Socket_getReadySocket`. Please double check if this makes sense. Or else, each packet handling may encounter 100ms delay when no socket is ready. This is critical for applications pursuing performance.

Another change is the timeout reducing from 1000ms to 10ms in the receive thread. It will make the receiving thread more responsive while has little impact on the CPU usage.

Here is the result with modified timeout in sample code. [mqtt_sample.txt](https://github.com/eclipse/paho.mqtt.c/files/14532898/mqtt_sample.txt)
``` text
====== Creating MQTT client ======
====== creating MQTT connect ======
====== Successful connection ====== 
====== Connect Time elpased is 1 ms ======

====== Creating MQTT client ======
====== creating MQTT connect ======
====== Successful connection ====== 
====== Connect Time elpased is 11 ms ======

====== Creating MQTT client ======
====== creating MQTT connect ======
====== Successful connection ====== 
====== Connect Time elpased is 9 ms ======
```

### DIFF
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -1357,7 +1357,10 @@ static int MQTTAsync_processCommand(void)
 			make sure we check for writeability as well as readability, otherwise we wait around longer than we need to
 			in Socket_getReadySocket() */
 			if (rc == EINPROGRESS)
+			{
 				Socket_addPendingWrite(command->client->c->net.socket);
+				rc = MQTTAsync_connecting(command->client);
+			}
 		}
 	}
 	else if (command->command.type == SUBSCRIBE)
@@ -2031,7 +2034,7 @@ thread_return_type WINAPI MQTTAsync_receiveThread(void* n)
 
 		if (sock == 0)
 			continue;
-		timeout = 1000L;
+		timeout = 10L;
 
 		/* find client corresponding to socket */
 		if (ListFindItem(MQTTAsync_handles, &sock, clientSockCompare) == NULL)
@@ -3017,8 +3020,6 @@ static MQTTPacket* MQTTAsync_cycle(SOCKET* sock, unsigned long timeout, int* rc)
 		MQTTAsync_lock_mutex(mqttasync_mutex);
 		should_stop = MQTTAsync_tostop;
 		MQTTAsync_unlock_mutex(mqttasync_mutex);
-		if (!should_stop && *sock == 0 && (timeout > 0L))
-			MQTTAsync_sleep(100L);
 #if defined(OPENSSL)
 	}
 #endif
@@ -3032,7 +3033,7 @@ static MQTTPacket* MQTTAsync_cycle(SOCKET* sock, unsigned long timeout, int* rc)
 		{
 			Log(TRACE_MINIMUM, -1, "m->c->connect_state = %d", m->c->connect_state);
 			if (m->c->connect_state == TCP_IN_PROGRESS || m->c->connect_state == SSL_IN_PROGRESS || m->c->connect_state == WEBSOCKET_IN_PROGRESS)
-				*rc = MQTTAsync_connecting(m);
+				;
 			else
 				pack = MQTTPacket_Factory(m->c->MQTTVersion, &m->c->net, rc);
 			if (m->c->connect_state == WAIT_FOR_CONNACK && *rc == SOCKET_ERROR)


TiejunZhou-CN:
@icraggs , just so you know, I'm in @ericwol-msft 's team. We are doing evaluation of open source Rust MQTT clients.

src/MQTTAsyncUtils.c
```diff
@@ -2031,7 +2034,7 @@ thread_return_type WINAPI MQTTAsync_receiveThread(void* n)
 
 		if (sock == 0)
 			continue;
-		timeout = 1000L;
+		timeout = 10L;
```
bartbizon, line 2037: I don't think this modification is correct.
The first time the function is entered, the call to MQTTAsync_cycle(&sock, timeout, &rc); uses timeout = 10.
If this loop has to be executed the second time around, then timeout is increased to 1000.
The question really is why was 1000 picked?

Let's consider the following: 
1. What is really this timeout value? 
Feel free to look at static MQTTPacket* MQTTAsync_cycle(SOCKET* sock, unsigned long timeout, int* rc).  
It's used in the Socket_getReadySocket(0, (int)timeout, socket_mutex, &rc1); call.
2. In the body of SOCKET Socket_getReadySocket(int more_work, int timeout, mutex_type mutex, int* rc), timeout is used to set a local variable timeout_ms, which as a good programming practice is initialized at the top of the function (happens to be 1000).
3.  Now, there are two versions of the SOCKET Socket_getReadySocket(int more_work, int timeout, mutex_type mutex, int* rc) function.  Although they are different, they seem to imply that a slowdown in operation might be desired if getting a socket did not succeed the first time around.

TiejunZhou-CN: I agree this change is not the best practice. This can be considered to be a temporary fix. The root cause of connection delay is that, not all socket operations are done in Socket_getReadySocket(). TCP connection is established in send thread. Even when it is completed, receive thread will continue wait until timeout. The best solution to avoid unexpected timeout for TCP connection is to move that logic (TCP connection) to receive Thread.

JinyongJeong-Riibotics:
Any update for this issue? I have same issue with this issue.

stitch3210:
I'd also like to upvote this - it seems it's not being picked up at all.
Our application takes a full 12 seconds at startup just unnecessarily waiting on these timeouts, because we need a handful of connections in a row. It should be 100ms at most.

src/MQTTAsyncUtils.c
fpagliughi: Yes. I'm looking into some of these performance issues now, too.

From a first look, it appears that the read thread is blocked in a poll() for the duration of this timeout. The poll is very responsive for the socket it is monitoring, but when you want to connect (and thus add a new socket to monitor), it can only happen after the poll() times out and gathers the new list of sockets to watch.

So, this the length of this timeout dictates the pacing for adding connections.

What we really want is a signal to exit the poll immediately when we want to add a new connection. This would be trivial on *nix systems (eventfd, mq, etc) but I'm not sure how to do it on Windows.

So, agreed, as a quick fix, we need to better manage the timeout.

To me, even the initial 10ms timeout is bad. You enter the thread initially because a connection is pending; any wait is an artificial delay. Better to loop very fast until that connection attempt completes.

After that, a 1sec delay is obviously too long. But spinning quickly and unnecessarily can eat up CPU, especially on a small board (Beaglebone, RPi, etc). Perhaps we should keep the timeout low (5-10ms) while new connections are being made, then slowly back off to a reasonable value, like 50-100ms when we achieve more of a steady-state?

fpagliughi: I played with this a little looking for a quick fix that might be beneficial with out significantly affecting the CPU usage in steady state.

It looks like we can make a nice improvement with a dynamic timeout:
- Start at 0ms. That appears to allow us to get the initial connection without sleeping
- Back off to 10ms at 1ms increments until socket activity is detected
- After socket activity starts, back off to 100ms in 10 ms increments, then keep a 100ms timeout
- **BUT** if the number of clients/sockets changes (from the application connecting or disconnecting), drop back down to 0ms and start the backoff again. 

To start up an app and make 5 connections to local host looks consistently like this on my dev machine:
```
Connect: 0.57ms, 10.15ms, 1.06ms, 1.09ms, 1.07ms 
```

To make 5 connections to localhost, publish a 1kB message to each at QoS=1, then disconnect them all takes ~15ms. Total. And that is waiting for the ACKs for each before doing the next.

I'll get a PR up shortly.

icraggs:
Somehow I missed this, but I think the best way is for the send thread to be able to interrupt the poll() call when some activity arrives as the poll is already waiting. 

An eventfd (Linux only), or socketpair (Linux, MacOS only), and some other solution for Windows.

icraggs:
Thanks for the suggestion. I've implemented my preferred solution in https://github.com/eclipse-paho/paho.mqtt.c/pull/1643

Issue #1464, opened 2024-04-09
IpV6 link local addresses resolve but don't connect


rdaum:
**Describe the bug**

ipv6 link local style addresses don't work.

paho.mqtt.c successfully _resolves_ addresses of the form

`tcp://[xxxx::xxxx:xxxx:xxxx%eth0]:1883`

But when it comes to connect, the complete results from `getaddrinfo` are not copied over into the `addrinfo`, as per line 1130 of `Socket.c`:

```c
		if (res->ai_family == AF_INET6)
		{
			address6.sin6_port = htons(port);
			address6.sin6_family = family = AF_INET6;
			memcpy(&address6.sin6_addr, &((struct sockaddr_in6*)(res->ai_addr))->sin6_addr, sizeof(address6.sin6_addr));
```		

This only copies over the address, but not the `sin6_scope_id` and `sin6_flowinfo`, so the complete link local address cannot connect.

Adding these two lines (or cleaner variant thereof) led to success for me:

```c
			memcpy(&address6.sin6_scope_id, &((struct sockaddr_in6*)(res->ai_addr))->sin6_scope_id, sizeof(address6.sin6_scope_id));
			memcpy(&address6.sin6_flowinfo, &((struct sockaddr_in6*)(res->ai_addr))->sin6_flowinfo, sizeof(address6.sin6_flowinfo));
		}
```

**To Reproduce**

Attempt to connect to an ipv6 link local address to a mqtt broker. It will fail at connect time. Adding the two lines listed above makes it succeed.

**Expected behavior**

Paho should be able to connect to a broker using such an address.

** Environment (please complete the following information):**
 - Linux
 - paho.mqtt.c checked out @ commit hash 6b1e202a701ffcdaa277b5644ed291287a70a7aa

rdaum:
Fix at https://github.com/eclipse/paho.mqtt.c/compare/master...rdaum:paho.mqtt.c:master

I'd create a PR, but I'm not going to go register and sign the Eclipse contributors agreement just for this.  Please apply it yourselves.

jumoog:
I made a PR.. #1466 I had to reset the author to pass eca check. I hope thats fine for everybody.

PR #1466, opened 2024-04-10
Fix IpV6 link local address connections

jumoog:
Fix IpV6 link local address connections

Paho currently successfully resolves IPv6 link-local addresses (of the
form `tcp://[xxxx::xxxx:xxxx:xxxx%eth0]:1883`). But it does not copy
the resolved link information over when attempting to make the
connection.

This is a 3-line fix.

See: https://github.com/eclipse/paho.mqtt.c/issues/1464


rdaum:
What needs to happen to get this change landed and released? We're running (@ my employer) with a local vendored fork (+ Rust bindings) and would prefer to be using upstream.

icraggs:
Thank you!

PR #1475, opened 2024-04-26
Add public key pinning

DBL2017:
Added a configuration item called “publicKey” to represent the server’s public key, which is used for pinning the server certificate with a public key.


fpagliughi:
Actually, this might be nice to have.

But, unfortunately, one thing you missed is how the library handles modifications to the public data structures, particularly the options like `MQTTAsync_SSLOptions`.



Basically, you should preserve the pre-existing binary layout; so only add new elements to the end of the struct. Then update the structure version number, and only access the new elements in the code if the in-memory struct has a version greater-than or equal to the version where you added the new elements.

But be careful when doing this. The current version number of the structs is hard-coded at a few places in the code. This should probably be fixed with some `#define` constants, but at the moment this is the way.

There's a similar discussion of this in another pending PR:
https://github.com/eclipse/paho.mqtt.c/pull/1463#issuecomment-2061196076

PR #1484, opened 2024-05-28
Fix usage of realloc

haraldreif:
When realloc returns `NULL`, the memory of the argument `ptr` remains allocated. Thus overwriting the variable that holds this argument directly with the return value would lead to memory leaks as the allocated address is overwritten with `NULL`. To fix this, the return value is buffered and checked for `NULL` before overwriting the address variable.


icraggs:
Thanks.

PR #1486, opened 2024-05-28
Fix compiler warnings

haraldreif:
There are several locations where a local variable in a scope hides a local variable in a superior scope. Some were different variables: they have been renamed. Some were redefinitions of an existing variable: the redefinition has been replaced with a reuse of the existing variable. Additionally, some uninitialized memory warnings have been resolved.


icraggs:
Thanks.

PR #1498, opened 2024-07-03
Zero out MQTTProperty before reading

fpagliughi:
This is a small nit...

It would be nice to zero out the `MQTTProperty` union/value before reading fields into it. That would leave any unused space predictably set to zero.

On advantage is that if you accidentally read an 8 or 16-bit value out using a larger field size (i.e. read as a u32) you still get the correct numeric value.

(Wonder what made me think to do this).

### DIFF
--- a/src/MQTTProperties.c
+++ b/src/MQTTProperties.c
@@ -240,6 +240,7 @@ int MQTTProperty_read(MQTTProperty* prop, char** pptr, char* enddata)
   int type = -1,
     len = -1;
 
+  memset(prop, 0, sizeof(MQTTProperty));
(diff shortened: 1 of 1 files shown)


jumoog:
How will this impact `PAHO_HIGH_PERFORMANCE`?

fpagliughi:
It is just an unconditional zero of memory for a tiny (~40byte) data struct. The performance hit in the overall read function is minimal, and that is overall likely undetectable in the overall read of a packet.

Though I definitely want to start giving a closer look to the performance characteristics of this library, we will need to look more at thread synchronization and things like that.

PR #1504, opened 2024-07-20
Small cleanup of OpenSSL/LibreSSL CMake

fpagliughi:
This implements the fix for the SSL CMake issue I commented on in #1451, [here](https://github.com/eclipse/paho.mqtt.c/pull/1451#issuecomment-2240902289)

@marcfir

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,8 +25,8 @@ project("Eclipse Paho C"
 message(STATUS "CMake version: " ${CMAKE_VERSION})
 message(STATUS "CMake system name: " ${CMAKE_SYSTEM_NAME})
 
-set(CMAKE_SCRIPTS "${CMAKE_SOURCE_DIR}/cmake")
-set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules")
+set(CMAKE_SCRIPTS "${PROJECT_SOURCE_DIR}/cmake")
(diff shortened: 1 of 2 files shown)


fpagliughi:
This also changes the CMake output to look a little more consistent (nicer, in my opinion):
```
-- CMake system name: Linux
-- Timestamp is 2024-07-20T04:11:46Z
-- Found OpenSSL: /usr/lib/x86_64-linux-gnu/libcrypto.so (found version "3.0.2")  
-- Using OpenSSL with headers at /usr/include, lib /usr/lib/x86_64-linux-gnu/libssl.so
-- Configuring done
```

fpagliughi:
I also just pushed a 2nd commit into the PR.

Using `CMAKE_SOURCE_DIR` is not correct to find files relative to the project's source tree. It is the top-level directory in the build, and thus will change if the project is bring built from a subdirectory, such as using a Git submodule. (Thus it will be the directory of the parent project, not the Paho C project directory).

Using `PROJECT_SOURCE_DIR` will keep it pinned to the Paho C sources even if built as a subdirectory. (Using `CMAKE_CURRENT_SOURCE_DIR` would have worked here as well).

icraggs:
I reverted this PR because it still didn't work for the MacOS and LibreSSL builds.

fpagliughi:
OK, I will have another look at it and try again.

fpagliughi:
@icraggs Apologies, that was a dumb mistake! ...which just so happened to work on my machines. I put up a new, fixed, PR #1550 , and it is passing CI.

Issue #864, opened 2020-04-27
Add support for Unix-domain sockets


fpagliughi:
I'd like to add support for Unix-domain sockets. 

These satisfy the MQTT v5 transport requirement for "ordered, lossless, [bidirectional] stream of bytes", but on a local machine, they provide higher throughput with lower latency and increased security in comparison to TCP.

A great use case for MQTT is providing an internal message bus on a remote device or computer, in which a large percentage of the messages are simply passed around inside the local machine. The use of MQTT for this is particularly helpful in IoT gateways where some subset of the messages are then bridged to a cloud server in a more traditional MQTT usage. But using UNIX sockets for the internal communications would make them faster and safer.

I have long rallied for this feature to be added to Mosquitto based on the performance metrics, to no avail. But Roger finally relented based on the security argument:
> The main reason that I changed my mind on this was the idea of having localhost only listener, where the access control was through the socket permissions. It's much easier to isolate a trusted listener that way.
https://github.com/eclipse/mosquitto/pull/467

I would suggest for the Paho Client(s) we use a URL in the form:
```
unix://path/to/socket
```

whereas if it's an absolute path, use the additional slash:
```
unix:///absolute/path/to/socket
```
And my biggest hope is that this is actually pretty trivial to implement in the client. Once opened, a Unix socket should mostly just behave as a socket.

icraggs:
I thought about this a long time ago, and I seem to remember implementing at least a trial.  I think, one of the reasons that it hasn't been done so far is that it isn't quite as straightforward as one would hope.  I would be happy to be proved wrong in that, but I wouldn't be surprised either way.

When you say "I'd like to add" ... does that mean you are considering a PR? :-)

fpagliughi:
You might have noticed my "once opened" caveat. Sometimes permissions and security issues trip you up to make a connection, but they're usually well behaved once running. But I was throwing this out to see if anyone could think of any other problems.

And I suppose that I am "considering" a PR. But I have a release of the Rust library to push out this week, then the C++ service release, and then a bigger update on the Rust lib to get it up to "modern" Rust standards.

So if anyone beat me to this, I wouldn't be (a) surprised, or (b) offended.

pulquero:
Came across this for the python client https://pypi.org/project/paho-socket/. Might be a stop-gap for some of us while we are waiting on this issue.

vjardinMisc:
Hi,

[...]

Issue #1376, opened 2023-07-07
Libraries having "-static" prefix are compiled with "Multi-treaded DLL" option on Windows


Alexk12:
**Describe the bug**
Libraries having "-static" prefix are compiled with "Multi-treaded DLL" and "Multi-treaded Debug DLL" option on Windows.

This actually means "Dynamic" libraries requiring run-time DLLs.
Not static.

As a result, it is impossible to link a C++ project compiled with /MT or /MTd options with Paho C library (paho-mqtt3as-static.lib).

icraggs:
Are you building the library yourself? If so, how are you doing it - with CMake? What options?

Alexk12:
> Are you building the library yourself? If so, how are you doing it - with CMake? What options?

Building Paho C:

call vc_env.bat
cd .\paho.mqtt.c
cmake -G "Visual Studio 16 2019" -A x64 -Bbuild -H. -DPAHO_WITH_SSL=TRUE -DPAHO_BUILD_SHARED=FALSE -DPAHO_BUILD_STATIC=TRUE -DCMAKE_INSTALL_PREFIX=C:\mqtt\paho-c -DLIB_EAY_DEBUG=C:/OpenSSL/lib/VC/static/libcrypto64MTd.lib -DLIB_EAY_RELEASE=C:/OpenSSL/lib/VC/static/libcrypto64MT.lib -DSSL_EAY_DEBUG=C:/OpenSSL/lib/VC/static/libssl64MTd.lib -DSSL_EAY_RELEASE=C:/OpenSSL/lib/VC/static/libssl64MT.lib
cmake --build build/ --target install
cd ..

Resulting "Eclipse Paho C.sln" includes project for Paho C static library, this project uses "Multi-theaded Debug DLL" option instead of "Multi-theaded Debug" and "Multi-theaded DLL" option instead of "Multi-theaded".

Building Paho CPP:

call vc_env.bat
cd .\paho.mqtt.cpp
cmake -G "Visual Studio 16 2019" -A x64 -Bbuild -H. -DPAHO_WITH_SSL=ON -DCMAKE_PREFIX_PATH=C:\mqtt\paho-c
cd ..

OS: Windows
Versions:
Microsoft Visual Studio Community 2019 Version 16.11.27
cmake version 3.20.21032501-MSVC_2
paho.mqtt.c 1.3.12
paho.mqtt.cpp 1.2.0

icraggs:
[...]

PR #1552, opened 2025-01-02
Add support for Unix sockets

fpagliughi:
Adds support for UNIX-domain sockets on *nix platforms.
It implements issue #864.

This allows clients to communicate with an MQTT server over UNIX-domain sockets on *nix systems. (Not on Windows). This is opt-in with a CMake build option, `PAHO_WITH_UNIX_SOCKETS`.

This can increase performance and reduce latency since UNIX sockets have a much simpler network stack than TCP. But perhaps more importantly, it provides increased security with more fine-grained control when clients reside on the same host as the broker. Access to the broker can be done using file permissions (with groups, owners, etc).

The clients can request a socket of this type using a URI of `unix:///path/to/socket/file`. The socket file is created by the server when it binds and listens on the UNIX socket.

The implementation here is fairly trivial. The library must now recognize the `unix://` scheme in the URI, then create and connect to a socket using a *sockaddr_un* socket address containing the path to the socket file.

After that all communications and closing the socket are handled by the same code that manages TCP connections. A socket is a socket.

This was tested on Linux and macOS with the _mosquitto_ server. By definition it requires a broker running on the same host whereby it communicates with the clients using the local file system.

The mosquitto configuration file can be as simple as:
```
allow_anonymous true    
listener 0 /tmp/mosquitto.sock
```

This PR also updates several of the simple sample apps to take a URI at the command line, allowing them to be used to try this out, like:
```
MQTTAsync_publish unix:///tmp/mosquitto.sock
```

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -329,6 +329,9 @@ int MQTTAsync_createWithOptions(MQTTAsync* handle, const char* serverURI, const 
 	{
 		if (strncmp(URI_TCP, serverURI, strlen(URI_TCP)) != 0
 		 && strncmp(URI_MQTT, serverURI, strlen(URI_MQTT)) != 0
+#if defined(UNIXSOCK)
+		 && strncmp(URI_UNIX, serverURI, strlen(URI_UNIX)) != 0
+#endif
 		 && strncmp(URI_WS, serverURI, strlen(URI_WS)) != 0
 #if defined(OPENSSL)
 		 && strncmp(URI_SSL, serverURI, strlen(URI_SSL)) != 0
@@ -384,6 +387,13 @@ int MQTTAsync_createWithOptions(MQTTAsync* handle, const char* serverURI, const 
 		serverURI += strlen(URI_TCP);
 	else if (strncmp(URI_MQTT, serverURI, strlen(URI_MQTT)) == 0)
 		serverURI += strlen(URI_MQTT);
+#if defined(UNIXSOCK)
+	else if (strncmp(URI_UNIX, serverURI, strlen(URI_UNIX)) == 0)
+	{
+		serverURI += strlen(URI_UNIX);
+		m->unixsock = 1;
+	}
+#endif
 	else if (strncmp(URI_WS, serverURI, strlen(URI_WS)) == 0)
 	{
 		serverURI += strlen(URI_WS);
--- a/src/MQTTAsyncUtils.h
+++ b/src/MQTTAsyncUtils.h
@@ -24,6 +24,7 @@
 #define URI_MQTT "mqtt://"
 #define URI_WS   "ws://"
 #define URI_WSS  "wss://"
+#define URI_UNIX "unix://"
 
 enum MQTTAsync_threadStates
 {
@@ -89,6 +90,7 @@ typedef struct
 typedef struct MQTTAsync_struct
 {
 	char* serverURI;
+	int unixsock;
 	int ssl;
 	int websocket;
 	Clients* c;
--- a/src/MQTTClient.c
+++ b/src/MQTTClient.c
@@ -80,6 +80,7 @@
 #define URI_MQTT "mqtt://"
 #define URI_WS   "ws://"
 #define URI_WSS  "wss://"
+#define URI_UNIX "unix://"
 
 #include "VersionInfo.h"
 #include "WebSocket.h"
@@ -294,6 +295,7 @@ typedef struct
 {
 	char* serverURI;
 	const char* currentServerURI; /* when using HA options, set the currently used serverURI */
+	int unixsock;
 #if defined(OPENSSL)
 	int ssl;
 #endif
@@ -407,6 +409,9 @@ int MQTTClient_createWithOptions(MQTTClient* handle, const char* serverURI, cons
          && strncmp(URI_MQTTS, serverURI, strlen(URI_MQTTS)) != 0
 		 && strncmp(URI_WSS, serverURI, strlen(URI_WSS)) != 0
 #endif
+#if defined(UNIXSOCK)
+		 && strncmp(URI_UNIX, serverURI, strlen(URI_UNIX)) != 0
+#endif
 			)
 		{
 			rc = MQTTCLIENT_BAD_PROTOCOL;
@@ -486,6 +491,13 @@ int MQTTClient_createWithOptions(MQTTClient* handle, const char* serverURI, cons
 		goto exit;
 #endif
 	}
+#if defined(UNIXSOCK)
+	else if (strncmp(URI_UNIX, serverURI, strlen(URI_UNIX)) == 0)
+	{
+		serverURI += strlen(URI_UNIX);
+		m->unixsock = 1;
+	}
+#endif
 	m->serverURI = MQTTStrdup(serverURI);
 	ListAppend(handles, m, sizeof(MQTTClients));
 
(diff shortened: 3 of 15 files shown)


src/MQTTAsyncUtils.h
```diff
@@ -25,6 +25,10 @@
 #define URI_WS   "ws://"
 #define URI_WSS  "wss://"
 
+#if defined(UNIXSOCK)
```
vjardin: This #if defined() is not needed, let's keep the code simple.

src/MQTTClient.c
```diff
@@ -81,6 +81,10 @@
 #define URI_WS   "ws://"
 #define URI_WSS  "wss://"
 
+#if defined(UNIXSOCK)
+#define URI_UNIX "unix://"
```
vjardin, line 83: Same

src/MQTTProtocolOut.c
```diff
@@ -318,6 +318,12 @@ int MQTTProtocol_connect(const char* ip_address, Clients* aClient, int websocket
 		rc = Socket_new(aClient->net.https_proxy, addr_len, port, &(aClient->net.socket));
 #endif
 	}
+#endif
+#if defined(UNIXSOCK)
+	else if (unixsock) {
+		addr_len = strlen(ip_address);
+		rc = Socket_unix_new(ip_address, addr_len, &(aClient->net.socket));
```
vjardin: can you rename ip_address to addr since it can be an IP address or a path name to a unix socket ?

src/MQTTProtocolOut.c
```diff
@@ -211,18 +211,18 @@ int MQTTProtocol_setHTTPProxy(Clients* aClient, char* source, char** dest, char*
  */
 #if defined(OPENSSL)
 #if defined(__GNUC__) && defined(__linux__)
-int MQTTProtocol_connect(const char* ip_address, Clients* aClient, int ssl, int websocket, int MQTTVersion,
+int MQTTProtocol_connect(const char* ip_address, Clients* aClient, int unixsock, int ssl, int websocket, int MQTTVersion,
```
vjardin: `ip_address` is not just an IP address, it can be a path to a unix socket, please, can you rename it accordingly ?

src/MQTTAsync.c
```diff
@@ -329,6 +329,9 @@ int MQTTAsync_createWithOptions(MQTTAsync* handle, const char* serverURI, const
 	{
 		if (strncmp(URI_TCP, serverURI, strlen(URI_TCP)) != 0
 		 && strncmp(URI_MQTT, serverURI, strlen(URI_MQTT)) != 0
+#if defined(UNIXSOCK)
```
vjardin, line 332: you should avoid such `if defined()` in order to avoid code complexity. The error cases can be handled for WIN32 or non UNIXSOCK into the function `Socket_unix_new()` that can be available for all the cases, but that should return a runtime ERROR when Unixsock is not available. 

So the code will be easier to read.

src/MQTTAsync.c
```diff
@@ -384,6 +387,13 @@ int MQTTAsync_createWithOptions(MQTTAsync* handle, const char* serverURI, const
 		serverURI += strlen(URI_TCP);
 	else if (strncmp(URI_MQTT, serverURI, strlen(URI_MQTT)) == 0)
 		serverURI += strlen(URI_MQTT);
+#if defined(UNIXSOCK)
```
vjardin, line 390: same, remove `#if defined(UNIXSOCK)`

src/Socket.c
```diff
@@ -1356,6 +1361,58 @@ int Socket_new(const char* addr, size_t addr_len, int port, SOCKET* sock)
 	return rc;
 }
 
+#if defined(UNIXSOCK)
```
vjardin, line 1364: No, this `int Socket_unix_new(const char* addr, size_t addr_len, SOCKET* sock)` should be available for all OS ; but in case of WIN32, it should return a PAHO error or a SOCKET_ERROR with a log that Unix socket is not available.

Doing so, you'll remove many `#if defined(UNIXSOCK).

src/samples/MQTTClient_publish.c
```diff
@@ -34,7 +34,10 @@ int main(int argc, char* argv[])
     MQTTClient_deliveryToken token;
     int rc;
 
-    if ((rc = MQTTClient_create(&client, ADDRESS, CLIENTID,
+	const char* uri = (argc > 1) ? argv[1] : ADDRESS;
```
vjardin: ident issue.

src/samples/MQTTClient_publish_async.c
```diff
@@ -65,7 +65,10 @@ int main(int argc, char* argv[])
     MQTTClient_deliveryToken token;
     int rc;
 
-    if ((rc = MQTTClient_create(&client, ADDRESS, CLIENTID,
+	const char* uri = (argc > 1) ? argv[1] : ADDRESS;
```
vjardin: ident issue.

src/samples/MQTTAsync_subscribe.c
```diff
@@ -137,7 +137,10 @@ int main(int argc, char* argv[])
 	int rc;
 	int ch;
 
-	if ((rc = MQTTAsync_create(&client, ADDRESS, CLIENTID, MQTTCLIENT_PERSISTENCE_NONE, NULL))
+	const char* uri = (argc > 1) ? argv[1] : ADDRESS;
```
vjardin, line 140: ident issue

src/samples/MQTTAsync_publish.c
```diff
@@ -146,7 +146,10 @@ int main(int argc, char* argv[])
 	MQTTAsync_connectOptions conn_opts = MQTTAsync_connectOptions_initializer;
 	int rc;
 
-	if ((rc = MQTTAsync_create(&client, ADDRESS, CLIENTID, MQTTCLIENT_PERSISTENCE_NONE, NULL)) != MQTTASYNC_SUCCESS)
+	const char* uri = (argc > 1) ? argv[1] : ADDRESS;
```
vjardin, line 149: ident issue

src/Socket.c
```diff
@@ -1276,14 +1281,14 @@ int Socket_new(const char* addr, size_t addr_len, int port, SOCKET* sock)
...
-				}
-	#endif
+			if (1)
+			{
+				int optsend = 100; //2 * 1440;
+				printf("Setting optsend to %d\n", optsend);
+				if (setsockopt(*sock, SOL_SOCKET, SO_SNDBUF, (void*)&optsend, sizeof(optsend)) != 0)
+					Log(LOG_ERROR, -1, "Could not set SO_SNDBUF for socket %d", *sock);
+			}
+#endif
```
vjardin, line 1291: you are not changing any code -> ident issue. The diff should be empty.

vjardin (changes requested):
Thansk @fpagliughi ; it'll be a nice security and performance improvement. I took the liberty to review your commits.

src/MQTTAsyncUtils.h
fpagliughi: True enough.

src/Socket.c
fpagliughi: I disagree in the sense that the previous indenting was incorrect and misleading. Admittedly, this has nothing to do with the rest of the PR; it's a holdover from my initial attempt to modify `Socket_new()` for UNIX sockets, before deciding that adding a separate `Socket_unix_new()` would be more clear.

I'd prefer to keep it.

src/samples/MQTTClient_publish_async.c
fpagliughi: Argh. Yeah, you're correct. There is inconsistent formatting in this project. Why are the samples using spaces when the rest of the code is using tabs?

src/MQTTProtocolOut.c
fpagliughi: I was about to argue against doing this (not change too much), but the change is isolated, doesn't clash with any other names, and makes sense. Also, then, the function comments need to be updated.

src/MQTTAsync.c
fpagliughi: I disagree here. The conditional compilation ensures that `unix://` is reported as an unsupported protocol if support is not compiled into the library - just the same way it's done with  `ssl://`, `mqtts://`, and `wss://`.

src/Socket.c
fpagliughi: I agree with you that all the conditional compilation in these files makes it quite difficult to follow the flow through this code. But I think, at this point, it is probably be better to be consistent with existing conventions than start doing things differently for a small portion of the code. And, realistically, the inline code for this feature is trivial compared to the rest of what's there.

fpagliughi:
@vjardin Thanks so much for the review! So nice to have this quick feedback. I made some changes, and added my reasoning for not making other changes.

vjardin:
OK, LGTM.

Regaarding NOSIGPIPE, if it is not defined, it means that you should handle the SIGPIPE, I did not check if it is properly done. But in 2025, it should not be a topic !

fpagliughi:
For years I made the argument to implement UNIX sockets for MQTT to reduce latency in local systems, but that never seemed to sway people. The security argument gets people's attention.

But I just ran my first test, compiling this branch into the Paho Rust library and using it on an MQTT v5 RPC system that I have for one of my customers. Each service exports a _ping_ function to check that it's alive and responsive. The full round-trip is four publishes:
```
client [request] -> broker -> service [response] -> broker -> client
```
I just tried it to compare _localhost_ TCP to Unix sockets:
```
$ svcctl ping  -U tcp://localhost:1883                                   
Service responded in 43.247ms

$ svcctl ping -U unix:///tmp/mosquitto.sock
Service responded in 2.530ms
```

Using TCP takes 43ms... using UNIX sockets takes 2.5ms.

That's a 94% reduction in round-trip latency!!!

@icraggs :  I need this mergerd and released ASAP!!!    :-)

vjardin:
Same, this requirement of unix socket is for both security and performances.

icraggs:
Great! Does this change the external interfaces - as in could we have it in the next 1.3.n release?

I'm close to getting a 1.3.14 release out (I'm doing a proxy issue at the moment ) @fpagliughi could you take a look at this build issue to help me out? https://github.com/eclipse-paho/paho.mqtt.c/issues/1376

fpagliughi:
Hey @icraggs, no this doesn't change the external interfaces. When built in, it just allows the library to recognize a new URI scheme ("unix://") and handle it accordingly. I would love to get it into v1.3.x

I will squash the commits in this PR to make it easy to apply to both branches, and run some tests myself to merge and build in the `develop` branch.

fpagliughi:
Later today I can get a Windows machine running and test #1376

fpagliughi:
OK, I squashed this PR and then tried to cherry-pick the resulting commit into the `develop` branch. It didn't merge cleanly due to the choose-an-interface code from v1.4 in _Socket.c_ 

So I cleaned up the merge and created another pull request, #1553.

So **both** PRs can be landed into their target branches... this one into v1.4 and the other into develop.

icraggs:
I use the paho_c_pub and paho_c_sub samples a lot. Their interface is similar to mosquitto_pub and sub. You can try the unix sockets without any changes, for instance:

`build.paho/src/samples/paho_c_pub x --connection unix:///tmp/mosquitto.sock --trace protocol`

PR #1556, opened 2025-01-05
Build all the sample applications for the static library

fpagliughi:
I was testing the combination of builds across platforms (shared vs static) and noticed that we're only building a few of the sample apps for the static library, which is a little unfortunate, especially when only building the static lib. This builds the rest of them. 

Keeping with the existing conventions, each of the static executables gets a `_static` suffix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.swp
 *.pyc
 /build.paho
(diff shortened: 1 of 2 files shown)


icraggs:
Thanks Frank.

PR #1559, opened 2025-01-09
Removed unnecessary _WIN64 conditional checks

fpagliughi:
Minor nit to start some cleanup.... 

The `_WIN64` preprocessor macro is always defined by the Microsoft compiler when `_WIN32` is defined. So you don't need to check if one or the other is defined. You can just check for `_WIN32`.
```
#if defined(_WIN32)
    // Windows code (32 or 64-bit)
#else
    // Non-windows code (Linux, macOS, etc)
#endif
```

If you need code specific for 32-bit vs 64-bit, you need need to check for `_WIN64` first/else, or nest it, like:
```
#if defined(_WIN32)
    #if defined(_WIN64)
        // Windows 64-bit code
    #else
        // Windows 32-bit code
    #endif
#else
    // Non-windows code (Linux, macOS, etc)
#endif
```

but we're not making that distinction here. To just check if we're on Windows, checking for `_WIN32` is sufficient.

From the Microsoft documentation:

> - _WIN32 Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, or x64. Otherwise, undefined.
>
> - _WIN64 Defined as 1 when the compilation target is 64-bit ARM or x64. Otherwise, undefined.

https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -41,7 +41,7 @@
 
 #include <stdlib.h>
 #include <string.h>
-#if !defined(_WIN32) && !defined(_WIN64)
+#if !defined(_WIN32)
 	#include <sys/time.h>
 #else
 	#if defined(_MSC_VER) && _MSC_VER < 1900
@@ -105,7 +105,7 @@ void MQTTAsync_global_init(MQTTAsync_init_options* inits)
 #define min(a, b) (((a) < (b)) ? (a) : (b))
 #endif
 
-#if defined(WIN32) || defined(WIN64)
+#if defined(_WIN32)
 void MQTTAsync_init_rand(void)
 {
 	START_TIME_TYPE now = MQTTTime_start_clock();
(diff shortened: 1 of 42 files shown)


src/MQTTAsync.c
```diff
@@ -105,7 +105,7 @@ void MQTTAsync_global_init(MQTTAsync_init_options* inits)
 #define min(a, b) (((a) < (b)) ? (a) : (b))
 #endif
 
-#if defined(WIN32) || defined(WIN64)
+#if defined(WIN32)
```
jumoog: shouldnt be that `_WIN32` as well?

fpagliughi: Oh, yes. Did I miss one? Thanks.

fpagliughi: Fixed. Thanks, @jumoog !

jumoog (approved):

icraggs:
Got some spare time on your hands @fpagliughi ?!

Thanks

fpagliughi:
I did over the holidays!

There are several more partially finished branches in my personal repo for whenever I have a little more time!

Perhaps I should open some issues to discuss the ideas floating around, before I waste time on PRs that might not ever land.

PR #1569, opened 2025-01-30
Fix build on GNU Hurd

Roman3349:
This pull request fixes a build error on systems running the GNU Hurd kernel - for example, Debian GNU/Hurd.

Example of failed build: https://buildd.debian.org/status/package.php?p=paho.mqtt.c


icraggs:
Thanks. This commit has been merged into the develop and 1.4 branches.

PR #1574, opened 2025-02-11
Added SSLPluginInterface.h to the Makefile alogside other exported he…

borisBozicGet:
- Created and added SSLPluginInterface.h alongside other exported header files. 
- Added a callback function for handling the SSL object passed as an input parameter in the SSLSocket_Connect
function.

What does this functionality provide?
This functionality is essential in cybersecurity as it enables users to obtain an X509 Certificate from the SSL object and verify the device attempting to connect. Users can include the SSLPluginInterface.h file in their project and set the callback function "SSLPluginInterface_setcallback(int (callback)(SSL))" with their implementation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -305,6 +305,7 @@ install: build
 	$(INSTALL_DATA) ${srcdir}/MQTTReasonCodes.h $(DESTDIR)${includedir}
 	$(INSTALL_DATA) ${srcdir}/MQTTSubscribeOpts.h $(DESTDIR)${includedir}	
 	$(INSTALL_DATA) ${srcdir}/MQTTExportDeclarations.h $(DESTDIR)${includedir}
+	$(INSTALL_DATA) ${srcdir}/SSLPluginInterface.h $(DESTDIR)${includedir}
 	- $(INSTALL_DATA) doc/man/man1/paho_c_pub.1 $(DESTDIR)${man1dir}
 	- $(INSTALL_DATA) doc/man/man1/paho_c_sub.1 $(DESTDIR)${man1dir}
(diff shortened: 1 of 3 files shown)


icraggs:
I don't think this is the way I would want to allow people to get a pointer to the SSL object - it doesn't follow any of the API conventions but is instead a back door.

You would have to sign the Eclipse ECA for a pull request to be accepted (see the checks and the template for PRs.)

What exactly are you expecting people to do with the SSL object once they've got it? What we do already is:
- X509_check_host
- X509_check_ip_asc
on the peer certfificate if verify is used.

borisBozicGet:
Hi! Thank you for answering, ok then i will do that.
When you get access to the SSL object you can extract the "Certificate Template" extension. 
It helps in identifying the type of certificate that was issued based on the predefined templates available in the CA's infrastructure. This is useful in where different types of certificates are issued for different purposes for example user certificates, server certificates and other.
We need the functionalities to react to the Certificate properties, we would only need the X509 object, but we choose to use SSL object because it was more generic. 

We could change it to:
"SSLPluginInterface_setcallback(int (*callback)(X509*))", that would work for us, is this a good solution for this? Or is it something that could be implemented from you?

borisBozicGet:
@icraggs 
Hi!
if we would changed it to:
"SSLPluginInterface_setcallback(int (callback)(X509))"

Would that work? Or do you have a different proposal?

icraggs:
Yes I prefer the callback to return the OpenSSL object. I'll look into adding that capability into the 1.4.0 branch.

borisBozicGet:
@icraggs 
Thank you!

Issue #1396, opened 2023-09-18
IPv6 Support


matthiasklein:
We have a local MQTT broker which is only accessible via IPv6 link local, where the name resolution happens via mDNS/Avavi.

I debugged into the Socket_new() function and found that getaddrinfo() returns both an IPv4 and an IPv6 entry.
The "prefer ip4 addresses" section ensures that only IPv4 is tried, which fails completely. No connection possible.

Basically the same I see on an IPv6-only host, which also gets an IPv4 address during name resolution. The "prefer ip4 addresses" section again ensures that no connection is possible.

Other tools like mosquitto_pub or libcurl work fine in these environments.

From libcurl I know that it uses a "happy eyeballs" implementation: https://en.wikipedia.org/wiki/Happy_Eyeballs I.e. parallel connects with IPv4 and IPv6 ...

I think the manual preselection of IPv6 via a callback is not useful, because the client does not know when to preselect IPv6. In my opinion we should try to realize a "happy eyeballs" implementation. Or at least try all getaddrinfo() entries one after the other. Then at least a connection would be established at some point.

Assuming you would do the name resolution outside of the Paho library, and only pass an IP. Is it possible to pass an additional host name for the SSL check?

When is the release of version 1.4.0 with the IPv6 prefix callback planned?

icraggs:
The application would know at least as well as the library whether to prioritize IPv4 or IPv6, no? "Not useful" means the client library does it rather than the application. Multiple simultaneous connections feels like a lot of work, relatively. I'll think about it. I don't have a particular date for the 1.4 release but obviously adding more items will extend the schedule. Having said that, this implementation wouldn't change externals so could be added to 1.3.x. I could then remove the IPV4/IPV5 callback implementation.

Adding an additional (any number?) of hostnames for the SSL hostname check - another issue for that?

matthiasklein:
I have created an issue for the SSL hostname: https://github.com/eclipse/paho.mqtt.c/issues/1402

icraggs:
I checked the happy eyeballs method and that doesn't look like a good option to me. The winner of the connection is the one that connects first - which is entirely random. In the case of MQTT, the TCP connection could work but maybe the broker is only accepting MQTT connections on one version, so we would have to issue MQTT connects and see if those work. In parallel that would be quite complicated and open to surprising behaviour.

I think the current method in version 1.4, with the callback, is a good option.

[...]

PR #1577, opened 2025-02-19
Update getaddrinfo options to support IPv6 hostname resolution

cwardcode:
Updated the `hints` struct so that it passes in the `AI_ADDRCONFIG` flag for the `ai_flags` property, in support of issue #1396.

This will enable auto-translation for hostnames to their proper IP address, depending on the IP address version used by the host. From the `getaddrinfo(3)` manual page:

```
     ai_flags       The ai_flags field to which the hints parameter points shall be set to zero or be the bitwise-inclusive OR of one or more of the values AI_ADDRCONFIG, AI_ALL, AI_CANONNAME, AI_NUMERICHOST, AI_NUMERICSERV, AI_PASSIVE, AI_V4MAPPED,
                    AI_V4MAPPED_CFG, and AI_DEFAULT.

                    AI_ADDRCONFIG   If the AI_ADDRCONFIG bit is set, IPv4 addresses shall be returned only if an IPv4 address is configured on the local system, and IPv6 addresses shall be returned only if an IPv6 address is configured on the local
                                    system.
``` 

This change has been tested on a linux platform, and I can confirm that I can now connect to a dual-stack endpoint broker from a device that only has IPv6 configured. I've also ensured that if only IPv4 is enabled for the device, it works as well.

All existing tests pass with zero regressions.

### DIFF
--- a/src/Socket.c
+++ b/src/Socket.c
@@ -1146,7 +1146,7 @@ int Socket_new(const char* addr, size_t addr_len, int port, SOCKET* sock)
 	sa_family_t family = AF_INET;
 #endif
 	struct addrinfo *result = NULL;
-	struct addrinfo hints = {0, AF_UNSPEC, SOCK_STREAM, IPPROTO_TCP, 0, NULL, NULL, NULL};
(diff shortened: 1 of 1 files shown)


cwardcode:
It seems as if the ECA will not pick up my commited address and instead uses the github-private address. Will re-submit with a PR using my public email address instead.

PR #1579, opened 2025-02-20
Fix memleak which is reported by cppcheck.

Dts0:
On Linux this memory leak does not happen. But on some microcontrollers, leakage may occur when the memory is low.


icraggs:
Thanks. This is merged into the develop and 1.4 branches.

https://github.com/eclipse-paho/paho.mqtt.c/commit/3f5b987f3b809915fae4bfe6f99b6053ec2595d8

PR #1582, opened 2025-02-25
Implemented custom SNI server name

### LINKED ISSUE #1580: Custom SNI validation
From an issue on the Paho Rust repo (https://github.com/eclipse-paho/paho.mqtt.rust/issues/242):

> Allow passing a custom server name that will be used for the SNI (server name indication), instead of using the host name. Use case: host name of the connection (IP) is different than the certificate's SNI (device serial number). See [Node.js TLS servername parameter](https://nodejs.org/api/tls.html#tlsconnectoptions-callback) and https://github.com/eclipse-paho/paho.mqtt.c/issues/171 for how this could be implemented.

This makes sense to me and should be easy to implement. I can get a PR up by next week.

fpagliughi:
This implements #1580, to provide an optional, separate SNI server name.

It adds a new `const char* serverName` parameter to the SSL options. If it is non-NULL, then it is used for SNI validation, otherwise the host name is used. Thus the new capability is opt-in, and defaults to the previous implementation if not provided, and therefore is completely backward compatible.

### DIFF
--- a/src/MQTTAsync.c
+++ b/src/MQTTAsync.c
@@ -609,7 +609,7 @@ int MQTTAsync_connect(MQTTAsync handle, const MQTTAsync_connectOptions* options)
 	}
 	if (options->struct_version != 0 && options->ssl) /* check validity of SSL options structure */
 	{
-		if (strncmp(options->ssl->struct_id, "MQTS", 4) != 0 || options->ssl->struct_version < 0 || options->ssl->struct_version > 5)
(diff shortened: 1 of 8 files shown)


fpagliughi:
It took two tries, but this appears to be working now and passing all the unit tests.

The simplest test for this is to use the IP address in the URL and then set the `serverName` in the SSL options to the host name, and `verify` enabled like:
```
URL -> mqtts://5.196.78.28:8886
serverName -> test.mosquitto.org
verify -> 1
```

A thing to note is that this only allows for providing a single server name, even when using High Availability with multiple URLs. If you omit the server name, each host name will be derived from the URL, just like before, but if you do add a server name, all addresses will verify to that name.

PR #1594, opened 2025-04-05
Enable building with BoringSSL

lalten:
I'm writing Bazel build files for paho.mqtt.c and want to link against BoringSSL instead of OpenSSL or LibreSSL.

`BIO_f_base64` (https://docs.openssl.org/3.3/man3/BIO_f_base64/) is deprecated and can't easily be used when using BoringSSL via Bazel: 
https://github.com/google/boringssl/blob/a553a9d65aa4e4f28a6c3007db14e9605e7c3bde/include/openssl/bio.h#L788-L797
This PR updates src/Base64.c to use the recommended `EVP_EncodeBlock`/`EVP_DecodeBase64` functions instead, which are available.

The other change in src/MQTTAsyncUtils.c is analogous to https://github.com/eclipse-paho/paho.mqtt.c/issues/1426 - BoringSSL targets OpenSSL 1.1.0 but not all APIs are actually implemented (see https://github.com/google/boringssl/blob/main/PORTING.md)

### DIFF
--- a/src/Base64.c
+++ b/src/Base64.c
@@ -48,40 +48,45 @@ static b64_size_t Base64_encodeDecode(
 	b64_size_t ret = 0u;
 	if ( in_len > 0u )
 	{
-		int rv;
-		BIO *bio, *b64, *b_in, *b_out;
-
-		b64 = BIO_new(BIO_f_base64());
-		bio = BIO_new(BIO_s_mem());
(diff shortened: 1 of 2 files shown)


icraggs:
Merged into the develop and 1.4 branches. Thanks.

PR #1626, opened 2025-11-13
Document tls:// prefix

konstaty:
I've added the TLS prefix to the documentation as suggested in https://github.com/eclipse-paho/paho.mqtt.c/pull/1593#issuecomment-3189725299


icraggs:
Thank you. I've merged this into the develop and 1.4 branches

https://github.com/eclipse-paho/paho.mqtt.c/commit/e8b6b3742a21b28f2c7326a58cca6e914cd018a8

PR #1628, opened 2025-11-14
Fix ssl handshake want write

### LINKED ISSUE #1627: Improper handling of SSL_ERROR_WANT_READ/SSL_ERROR_WANT_WRITE
**Describe the bug**
SSL handshake fails intermittently on congested networks due to improper handling of `SSL_ERROR_WANT_WRITE`. When OpenSSL temporarily makes the socket non-blocking during the TLS handshake and the network TX buffer fills up, `SSL_connect()` returns `SSL_ERROR_WANT_WRITE` to indicate the caller must wait for the socket to become writable before retrying. However, the current code in `SSLSocket_connect()` treats `SSL_ERROR_WANT_WRITE` the same as `SSL_ERROR_WANT_READ` - it returns `TCPSOCKET_INTERRUPTED` without adding the socket to the pending-write poll set. This causes the MQTTClient background thread to spin-loop calling `SSL_connect()` repeatedly without waiting for socket writability, eventually exhausting the connection timeout and failing.

**To Reproduce**
This bug manifests on real networks with sustained congestion or limited bandwidth. It is difficult to reproduce on localhost due to the loopback interface's extremely low latency and unlimited throughput. It was identified on a production device on a customer's network, but can be replicated by using an LD_PRELOAD wrapper that mocks 'SSL_connect()' to artificially inject SSL_ERROR_WANT_WRITE.

1. **Create mock wrapper** (`ssl_mock_want_write.c`):
```c
#define _GNU_SOURCE
#include <dlfcn.h>
#include <openssl/ssl.h>
#include <openssl/err.h>
#include <stdio.h>
#include <unistd.h>

static int (*real_SSL_connect)(SSL *ssl) = NULL;
static int inject_count = 5;  // Inject WANT_WRITE first 5 times

int SSL_connect(SSL *ssl) {
    if (!real_SSL_connect)
        real_SSL_connect = dlsym(RTLD_NEXT, "SSL_connect");
    
    // Simulate congested network by delaying and forcing non-blocking behavior
    if (inject_count > 0) {
        inject_count--;
        usleep(200000);  // 200ms delay to simulate slow network
        fprintf(stderr, "MOCK: Forcing SSL_ERROR_WANT_WRITE (remaining: %d)\n", inject_count);
        // Let real SSL_connect run on small-buffer socket
        // The small buffers + delay will often naturally produce WANT_WRITE
    }
    
    return real_SSL_connect(ssl);
}
```

2. **Compile the wrapper**:
```bash
gcc -shared -fPIC -o libssl_mock.so ssl_mock_want_write.c -ldl
```

3. **Run test with unfixed library**:
```bash
# Set very small socket buffers to increase chance of WANT_WRITE
echo 4096 | sudo tee /proc/sys/net/core/wmem_default

# Run with mock
LD_PRELOAD=./libssl_mock.so ./mqtt_client_test
```

4. **Expected result**: Connection timeout after 30 seconds with logs showing repeated "SSLSocket error WANT_READ/WANT_WRITE"

5. **Run test with fixed library**: Connection succeeds even with injected delays

**Expected behavior**

According to OpenSSL documentation:

> **SSL_ERROR_WANT_WRITE:** The operation did not complete and can be retried later. The calling application should call `select()` or `poll()` to wait for the socket to become writable before calling `SSL_connect()` again.

When `SSL_ERROR_WANT_WRITE` occurs, the client should:
1. Add the socket to the pending-write poll set using `Socket_addPendingWrite(sock)`
2. Return `TCPSOCKET_INTERRUPTED` to signal retry needed
3. Background thread calls `Socket_getReadySocket()` which blocks on `poll/select` waiting for `POLLOUT`
4. When socket becomes writable, retry `SSL_connect()`
5. Handshake completes successfully

**Screenshots**

**Log files**
See attached log

- **OS:** Linux (embedded device)
- **Library Version:** paho.mqtt.c 1.3.13 (also affects latest master branch)
- **OpenSSL Version:** 1.1.1f
- **Network:** Cellular/WiFi with variable conditions
- **MQTT Client:** Synchronous API (`MQTTClient`)

**Root cause code (src/SSLSocket.c lines 778-781):**
```c
int error;
error = SSLSocket_error("SSL_connect", ssl, sock, rc, cb, u);
if (error == SSL_FATAL)
    rc = error;
if (error == SSL_ERROR_WANT_READ || error == SSL_ERROR_WANT_WRITE)
    rc = TCPSOCKET_INTERRUPTED;
```

**Problem:** Both `SSL_ERROR_WANT_READ` and `SSL_ERROR_WANT_WRITE` return `TCPSOCKET_INTERRUPTED`, but only `WANT_READ` requires waiting for readability. `WANT_WRITE` requires waiting for **writability**, which needs the socket added to the pending-write poll set.

**Proposed fix:**
Replace the if-statement with a switch statement that properly handles each error case:

```c
switch (error)
{
case SSL_FATAL:
    rc = error;
    Socket_clearPendingWrite(sock);
    break;
case SSL_ERROR_WANT_WRITE:
    /* SSL temporarily switched the socket non-blocking; wait for POLLOUT */
    Socket_addPendingWrite(sock);
    rc = TCPSOCKET_INTERRUPTED;
    break;
case SSL_ERROR_WANT_READ:
case SSL_ERROR_WANT_CONNECT:
case SSL_ERROR_WANT_ACCEPT:
case SSL_ERROR_WANT_X509_LOOKUP:
    Socket_clearPendingWrite(sock);
    rc = TCPSOCKET_INTERRUPTED;
    break;
default:
    Socket_clearPendingWrite(sock);
    rc = error;
    break;
}
```

Additionally, on successful handshake completion, clear the pending-write state:
```c
else
{
    /* Handshake finished, stop polling for write readiness */
    Socket_clearPendingWrite(sock);
    // ... existing certificate verification code
}
```

**Impact:**

This bug causes intermittent connection failures in production environments, particularly:
- Mobile/IoT devices on cellular networks
- Devices on congested WiFi networks
- Deployments during peak traffic periods
- Any scenario with limited network bandwidth

The fix implements the standard OpenSSL non-blocking I/O pattern and has zero downside - it only activates when `SSL_ERROR_WANT_WRITE` occurs.

spencerbug:


jumoog (approved):

icraggs:
Thanks. Trying to work out why the build checks didn't run.

icraggs:
Merged into the develop branch

 https://github.com/eclipse-paho/paho.mqtt.c/commit/3c7f634a307472b665307e73bbb52a96acc65164

Thanks!

PR #1635, opened 2025-12-25
Modify the conditions for successful execution of SSLSocket_setSocketForSSL()

youngbug:
The `SSLSocket_setSocketForSSL()` function returns 1 upon success, while calls to OpenSSL APIs within the function typically return 0 upon failure. However, OpenSSL API failures may also return other negative values. When an OpenSSL API call returns a value less than 0, using the condition `setSocketForSSLrc != 0` to check if `SSLSocket_setSocketForSSL()` succeeded will result in an error, potentially leading to serious errors in subsequent calls to `SSL_connect`.

### DIFF
--- a/src/MQTTAsyncUtils.c
+++ b/src/MQTTAsyncUtils.c
@@ -2949,7 +2949,7 @@ static int MQTTAsync_connecting(MQTTAsyncs* m)
 			setSocketForSSLrc = SSLSocket_setSocketForSSL(&m->c->net, m->c->sslopts,
 					serverURI, hostname_len);
 
-			if (setSocketForSSLrc != MQTTASYNC_SUCCESS)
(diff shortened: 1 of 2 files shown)


jumoog:
but `MQTTCLIENT_SUCCESS`  is `0`

icraggs:
@jumoog I think this is correct. The line containing MQTTCLIENT_SUCCESS is the one being removed.

jumoog:
What I mean is that == 0 or != 1 is logical the same

icraggs:
rc != 0 and rc == 1 are the same for 0 or 1, but not if the return code is negative, which it could be. The success return code from setSocketForSSLrc is 1, anything else is a failure.

icraggs:
I've merged these changes into the develop and 1.4 branches.

Issue #1608, opened 2025-07-01
CMake deprecation warning


fpagliughi:
The latest CMake v4.x is giving this deprecation warning when building Paho C:

```
  Error building Paho C library.

  --- stderr
  CMake Deprecation Warning at CMakeLists.txt:18 (cmake_minimum_required):
    Compatibility with CMake < 3.10 will be removed from a future version of
    CMake.

    Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
    to tell CMake that the project requires at least <min> but has been updated
    to work with policies introduced by <max> or earlier.
```

In the last update to this library, we bumped CMake to v3.5, but this is still too old for the latest build systems. The next release should bump it again to something >= v3.10.

IIRC, the was an issue last year that could have been easily solved with CMake v3.15 or above. I will look for that and see if it's worthwhile to bump it up to there.

Either way, I can investigate and put up a PR.

icraggs:
@fpagliughi any thoughts on this?

fpagliughi:
I think this is as easy as just bumping the minimum required CMake to something a little more recent. I've been scrolling through the CMake release notes to look for specific features that might be helpful, but nothing in particular jumps out. I do remember that 3.15 made writing some generator statements for the compilers much easier.

fpagliughi:
Apologies, I forgot to put up the PR for this. It was trivial; just bump the minimum CMake version in the top-level build file.

Pascal-Fremaux:
Warning: your warning issue is not related to cmake 4.x but to cmake 3.27 or 3.31 :
See https://github.com/Kitware/CMake/commit/84db8506ffc38381dd2502be41879ffe03d8e8e9
.. versionchanged:: 3.27
  Compatibility with versions of CMake older than 3.5 is deprecated.
=> warning if cmake_minimum < 3.5 (not <= 3.5)
.. versionchanged:: 3.31
 Compatibility with versions of CMake older than 3.10 is deprecated.
=> warning if cmake_minimum < 3.10

But for cmake 4.0, it seems this is not supported (no build) with cmake_minimum < 3.5 (OK above 3.5)
See https://fedoraproject.org/wiki/Changes/CMake4.0 
I did not go further for cmake 4.1.x or 4.2.x to see if they remofe support below 3.10.

fpagliughi:
[...]

PR #1637, opened 2026-01-10
Bumped minimum CMake to v3.12

fpagliughi:
Fix for #1608

This bumps the minimum version for CMake to get rid of the deprecation warning when using CMake v4.x.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 #     Genis Riera Perez - Add support for building debian package
 #*******************************************************************************/
 
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.15)
 
 project("Eclipse Paho C" 
   VERSION 1.3.15
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 #     Genis Riera Perez - Add support for building debian package
 #*******************************************************************************/
 
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.12)
 
 project("Eclipse Paho C" 
   VERSION 1.3.15


jumoog (approved):

icraggs:
Hi Frank. This update cause one of the AppVeyor runs to fail because it is the oldest Visual Studio environment which has an old version of CMake:

https://ci.appveyor.com/project/eclipsewebmaster/paho-mqtt-c/builds/53352392

It would be nice to still include this enviroment for testing, but I realize that it might be more trouble than it's worth. There's also the option to switch to Github hosted WIndows environments, which don't include anything that old.

I don't think the old environments ever catch any functional bugs, it's more about the build environment, which environment variables are available, which Windows API calls are supported on old versions, that sort of thing.

fpagliughi:
If I'm reading the AppVeyor file correctly, the oldest build environment is MSVS 2013. 

Apparently Microsoft stopped mainstream support for this product in 2019! And extended support ended two years ago. It is fully deprecated.

CMake < v3.10 is deprecated. (Which got us here)

I appreciate not wanting to break things for existing users, but I don't think we need to support tools that even the manufacturers don't support any longer. I doubt that a lot of people who haven't upgraded their compilers in a decade are keeping up with the latest library updates.

If someone is still using MSVC 2013, then Paho C v3.15 will still work fine for them.

We should keep moving forward.

fpagliughi:
Ah, well... I still totally believe my comment from above :-)

But I wrote it before even bothering to look at the AppVeyor error. It _does_ claim support for CMake v3.12. That would at least get us past the CMake deprecation, though it still prevents us from using some features of CMake that appeared later. 

So we can test out CMake v3.12. I assume it should work fine with the existing build. But at some point in the near future, we would likely need to bump it forward again.

fpagliughi:
OK. I moved it back to CMake v3.12, and that seems to satisfy the AppVeyor build. For now!

icraggs:
Ok, 3.12 works. Thanks!

One way to handle old stuff would be to have old version branches, but that would mean maintaing multiple streams, which I've tried to avoid because of the extra work involved. CMake deprecation will force us to move on at some point.

PR #1639, opened 2026-01-11
Consolidated "Event" thread signaling object.

### LINKED ISSUE #1638: Thread sync object cleanup
In an effort to clean up some of the base thread synchronization and improve performance, updates to the core threading functions can be done in two steps:

1. Consolidate the existing "event" signaling mechanisms into something simpler.
2. Replace some of the event signaling retry loops with condition variables to eliminate polling wait times.

This issue describes the first of these steps.

The library, in its current state, has two different thread signaling mechanisms: _semaphores_ and _conditions_. But these are _both_ being used as simple thread signals to indicate an event has occurred, similar to a binary semaphore.

This has created redundant code, which is made worse by the implementation differences for each target platform (Windows, Mac, and Linux/POSIX), and the need for conditional compilation scattered across the code base. In addition, there is a race condition in the Linux/Posix implementation, and some code paths in the implementations that aren't being used by the library.

The existing implementation can be significantly improved by replacing the two different sync objects with a single "event" object that can be simply and properly across the target platforms using a condition variable around a boolean flag indicating whether or not the object is currently "signaled".

This would reduce the code required in the thread module, eliminating some of the trickier platform-specific stuff, and a latent bug or two. And it would get rid of the need for the conditional compilation that exists in the rest of the code to both signal and wait for events.

fpagliughi:
This is a solution for #1638

The existing code uses both semaphores and conditions to create simple event signals (i.e. both implementations are essentially doing the same thing across platforms). 

This update combines the implementations into a single one, calling them "event" objects. 

- All *nix platforms (Linux, MacOS, Posix) share a simple implementation using condition variables
- On Windows they are implemented with Win32 Event objects
- All platform-specific code is contained in _Thread.c/.h_. 
    - The application code no longer needs to have separate paths for Windows to signal or wait for events.
    - Application code that was going directly to the OS APIs now uses the wrapper type
- Some possible fixes for bugs and platform weirdness from differences in semaphore behavior on different *nix operating systems, including a race condition on Linux.
 
Note that this does not try to improve performance or change the way the application signals events. It merely consolidates the code to do the signaling.


icraggs:
Thanks Frank. I fixed up the copyright statements - there must be a maximum of two years, the start and the end. So we just need to change the second year when updating a file.

I also had to add a WINAPI modifier to the line

`thread_return_type evt_secondary(void* n)`

in test/thread.c. Without that it failed to build on the Visual Studio 2019 build on AppVeyor. (See https://ci.appveyor.com/project/eclipsewebmaster/paho-mqtt-c/builds/53461610/job/prjp9bdrl4p13lqk)

Weirdly the other two AppVeyor build environments worked. Now I'm going to see about integrating these changes into the 1.4 branch.

PR #1642, opened 2026-02-07
Reduce latency on connect #1430

### LINKED ISSUE #1430: long connection latency when setting up multiple mqtt connections

icraggs:
Two aspects to this fix:

1. Interrupt the poll() call when new work comes along. The new work is adding or deleting a socket.
2. Reduce the poll() for read timeout if at least one of the sockets is in the middle of a non-blocking TCP connect as we are only checking for writeability, not waiting for readability.

Poiint 1 uses socketpair() which does not exist on Windows. So we have to develop an equivalent.

Needs cleaning up as well as the Windows implementation.


src/Socket.c
```diff
@@ -628,10 +670,42 @@ SOCKET Socket_getReadySocket(int more_work, int timeout, mutex_type mutex, int*
 			*rc = SOCKET_ERROR;
 			goto exit;
 		}
+		/* TODO: if there is a writeable socket in the connecting state (see isReady)
+		 * then set the timeout for the next poll to 0, so we can handle the connect
+		 * immediately */
```
fpagliughi: ^^^ Yes. Do this! 😄

src/Socket.c
```diff
@@ -628,10 +670,42 @@ SOCKET Socket_getReadySocket(int more_work, int timeout, mutex_type mutex, int*
...
+		mod_s.saved.cur_fd = 0;
+		while (mod_s.saved.cur_fd != -1)
+		{
+			if (isConnectReady(mod_s.saved.cur_fd))
+			{
+				printf("Socket %d is connecting, reducing timeout from %d, sockets %d\n", mod_s.saved.cur_fd, timeout_ms,
+					mod_s.saved.nfds);
+
+				timeout_ms = (mod_s.saved.nfds > 2) ? 10 : 10;
```
fpagliughi: Is this the place to do it?

If we simply set this to zero here, the dead sleep disappears from all the connect times. Otherwise we always hit this sleep for _every_ connect, even if the connection completes in <1ms.

So making this:
```
timeout_ms = 0;
```

works great over localhost on a fast machine. But would it peg the CPU indefinitely for a slow connection on a small board?

If so, could we set to zero here, then back it off slowly (in the read thread, maybe)?

fpagliughi: Setting this to zero gets my little speed test to this on a Linux desktop:

```
Create: 0.568, 0.022, 0.002, 0.001, 0.003, 
Connect: 0.527, 0.221, 0.156, 0.103, 0.093, 
Publish: 0.096, 0.054, 0.032, 0.034, 0.035, 
Disconnect: 0.046, 0.024, 0.025, 0.026, 0.023, 

Total: 2.162
```
 All times in ms.

That would get us down from 8500ms in the current implementation down to 2ms!!!

I could live with a 4000x performance increase.

icraggs: I'm still concerned that this could cause 100% CPU on a slow TCP connect, such as over a satellite link. The backoff approach could work here. If we set it to 0 for now, along with a cautionary comment, and see what happens when people use it, is probably my preferred option.

PR #1643, opened 2026-02-08
Some additional performance improvements for Issue #1430

fpagliughi:
These further reduce sleep wait times for #1430, with async clients.

Tested on Linux, and shows a marked improvement, particularly on fast networks, such as LANs and localhost. No  unintended consequences were noticed on some initial performance tests, but more testing is definitely required.


fpagliughi:
Rebased over the Windows update

icraggs:
Thanks Frank. I've made these changes but with comments and preserving some of the old statements for future reference.

PR #1645, opened 2026-02-11
Added a .clang-format file and bash script to run it 

fpagliughi:
This project has accumulated a number of different coding styles over that last few years (decades?). I personally keep forgetting to turn on tab indenting when I modify some files.

This could be helpful to make the code consistent and keep it that way. I tried to figure out and match the prevailing style, but feel free to make any changes to the way it formats the code. I don't care about any particular style. It would just be nice to keep it consistent, and to format any changes I make before sending a PR.

Note: I would recommend merging branches and PRs before doing a big reformat. And/or also reformatting the branches before attempting to merge. And I also recommend doing a single, isolated commit of just the reformat code.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,79 @@
+---
+Language: C
+
+# Indentation: tabs, one tab per indent level
+# The Emacs footer in source files specifies indent-tabs-mode: t
+UseTab: ForIndentation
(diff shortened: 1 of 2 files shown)


jumoog:
good idea

Issue #1654, opened 2026-03-03
CPACK variables are set even when PAHO_ENABLE_CPACK is turned off


fpagliughi:
An issue was raised in the Paho C++ repo:
https://github.com/eclipse-paho/paho.mqtt.cpp/issues/570

stating:

>Some CPACK variables are set even when the PAHO_ENABLE_CPACK option is turned off. This is a problem when the CMake project is included into another CMake project (as a sub-project).

I'll have a look here, verify, and fix if necessary.

fpagliughi:
The PR is up, #1665

PR #1665, opened 2026-03-30
Fixed CMake build so that CPack is only pulled in for PAHO_CPACK_ENAB…

fpagliughi:
Fix for #1654 

CPack was being enabled and some CPACK variables were set, even when PAHO_ENABLE_CPACK=OFF.

This PR:

- Moves `include(CPack)` and set(CPACK_PACKAGE_...)` into a conditional block
- Sets `PAHO_ENABLE_CPACK` if `PAHO_BUILD_DEB_PACKAGE` is set (since it's now required as the DEB builder uses CPack)


JeroenBrunnenkreef:
Any chance this will be merged in the next release?

PR #1672, opened 2026-05-14
Fix the QNX 8.0 build for Makefile and CMake

jclsn:
Fixes the QNX 8.0 build for Makefile and CMake. 

Supersedes https://github.com/eclipse-paho/paho.mqtt.c/pull/1602

Fixes the Make and CMake builds

Changes:

Code:
- SHA1.c: endian handling via gulliver.h
- Define AI_ADDRCONFIG=0, since it is not defined in all QNX versions

CMake:
- Add -D_QNX_SOURCE

Makefile:
- Removed -pthread, since is implied in QNX
- Added -lsocket, which is mandatory in QNX for networking
- Removed -ldl, since is not available in QNX
- defaults to qcc for QNX and gcc for Linux and Darwin in Makefile

Build instructions:

```sh
cmake -DCMAKE_SYSTEM_NAME=QNX -B build -S .
cmake --build build
``` 

or

```sh
make OS_TYPE=QNX
```

This builds with qcc by default, but you can specify CC=ntoaarch64-gcc
or CC=ntox86_64-gcc

- [x] QNX cross-compilation works with `qcc` and `ntoaarch64-gcc` and `ntox86_84-gcc`
- [x] Native QNX build works
- [x] Additionally verified correct behavior for the Linux build
- [x] Fixed some indendatation and const qualifier warnings, I was annoyed by. Can drop them if desired.


jclsn:
@icraggs Can you please have a look?

PR #1691, opened 2026-08-13
[persistence] Fix creation of absolute paths on windows

autoantwort:
Calling `_mkdir("C:")` yields an EACCES error. But making this call in the first place does not makes sense, since you can not create new drives by calling mkdir. Therefore we now skip to the first real folder on windows.


autoantwort:
The ci did not because: 
> This workflow run required approval but was not approved before it expired.

PR #1701, opened 2026-09-17
Use current binary directory for generated VersionInfo.h

CDSuadS:
Change the output path of the generated VersionInfo.h from `${CMAKE_BINARY_DIR}` to `${CMAKE_CURRENT_BINARY_DIR}`.
### **Motivation**
When Paho MQTT C is included as a subdirectory or through FetchContent in a larger CMake project, `${CMAKE_BINARY_DIR} `refers to the top-level project's binary directory.
As a result, VersionInfo.h is generated in the parent project's build directory. Since MQTTAsync.c and MQTTClient.c include this header, changes to the parent project's CMake configuration can cause Paho's object files to be considered out of date and rebuilt, even when the Paho source and compile commands have not changed.
Using `${CMAKE_CURRENT_BINARY_DIR}` keeps the generated header scoped to Paho's own build directory and avoids this unnecessary coupling with the parent project.
### **Testing**
Tested Paho MQTT C as a Git submodule included through FetchContent in a larger CMake project.

- Confirmed that the generated header is placed in Paho's build directory.
- Confirmed that modifying an unrelated parent-project CMakeLists.txt no longer causes Paho's source files to be recompiled unnecessarily.
- Confirmed that Paho builds successfully with the existing configuration.


CDMartinL (approved):
lgtm

